PR #2, opened 2020-05-20
Fix paths

mhsekhavat:


mhsekhavat:
@mhejrati I have pushed images and tested all dockerfiles except coral-dev-board. I will test it too.

mhsekhavat:
@mhejrati I have tested coral-dev-board too. All dockerfiles work with this PR. Please do a final review.

PR #3, opened 2020-05-22
Implement react,fastapi,zmq

mhsekhavat:
This pull request creates a FastAPI backend (replaces Flask), decouple from the react frontend. The frontend has a separate docker image and is served as static files in /srv/frontend.

In order to stream the cameras to frontend, it uses zmq so that the processing is done only once and each client-serving thread just subscribes to the zmq. It is more efficient than the previous implementation but requires further optimization using shared memory or other streaming alternatives.

Also I have added minimal scripts for running, building and releasing docker images in `tasks` folder. It uses `pyinvoke` library and is intended to be used by neuralet dev team only. In order to run commands like `invoke dockerhub.release`, run `pip install invoke`.


mhejrati:
Thanks @mhsekhavat 

I am testing and reviewing this now. Is it WIP or ready to be merged into master?

I have difficulty running it on the coral TPU, not sure what is causing that, looks like an issue with my storage attached to the board. I will test more and let you know. Did you already test it on Coral Devboard?

mhsekhavat:
@mhejrati I tested it on TPU before server upgrade. It used to work but now I am getting a devicemapper-related error on docker build.

mhsekhavat:
It will be ready to merge tomorrow. I will notify you once I have pushed the final changes. It works with x86_openvino, but weirdly it doesn't work on x86.

PR #9, opened 2020-05-29
change pipeline string to match working test

mdegans:
not sure if this will work immeditely, but we know these components are needed now


mhsekhavat:
Thanks

PR #10, opened 2020-05-29
Delete nms and tracker

alpha-carinae29:
This will delete some of post processing steps since they are irrelevant duo to adding high accuracy pedestrian detectors.


JsonSadler:
What was the harm? :D
I think keeping them is beneficial as some one may needed to use simpler/weaker detectors on less powerful platforms.
Instead, we can add a configuration flag for enable/disable it. 
What do you think? @mhejrati

mrn-mln:
@alpha-carinae29 I've implemented a new tracker and remove the centroid tracker class.

PR #17, opened 2020-06-05
Gst merge

mdegans:
So, I got it working. With file sources there is a "bug" where it'll go as fast as it can, but it should work for live sources.
deepstream working

Next I will modify the osd code to get rid of those huge labels and add tegra support, but I think i'm signing off for the week now that this is working.


mdegans:
I made some changes to the entrypoint. Feel free to accept or reject as you feel appropriate @mhsekhavat

We proabably also discuss expected behavior when the gstreamer process is finished. See you on slack!

PR #23, opened 2020-06-22
Move Dashboard into separate component

mats-claassen:
Depends on `gstreamer` branch to be merged.

The WebGUI has been moved to a separate component built with `run-frontend.Dockerfile`.
The frontend has a config file  which tells it where to look for the processor node.
This is a temporal workaround as this should be a setting in the Dashboard.

Also moved the `visualization_utils.py` file from the `ui` folder to `libs` as it is not used by the frontend but by the processor.

TODO: (maybe in separate PRs?)
 
 - [ ] Support pushing analytics to some Analytics storage
 - [ ] Fetch analytics from Dashboard
 - [ ] Configure IP of processor node(s) from Dashboard settings
  
For the analytics parts we might have to define an interface first.


neuralet-distancing.py
```diff
@@ -18,9 +18,9 @@ def start_engine(config, video_path):
 
 
 def start_web_gui(config):
```
JsonSadler, line 20: I suggest renaming this too, to something indicating that it starts the processor's api

api/processor_api.py
```diff
@@ -0,0 +1,41 @@
+import time
+from fastapi import FastAPI
+from fastapi.staticfiles import StaticFiles
+from fastapi.responses import FileResponse, StreamingResponse
+import uvicorn
+import os
+
+class ProcessorAPI:
+    """
+    The Webgui object implements a fastapi application and acts as an interface for users.
```
JsonSadler, line 10: Docstrings need update

api/processor_api.py
```diff
@@ -0,0 +1,41 @@
...
+
+        if os.environ.get('DEV_ALLOW_ALL_ORIGINS', False):
+            # This option allows React development server (which is served on another port, like 3000) to proxy requests
+            # to this server.
+            # WARNING: read this before enabling it in your environment:
+            # https://medium.com/@stestagg/stealing-secrets-from-developers-using-websockets-254f98d577a0
+            from fastapi.middleware.cors import CORSMiddleware
+            app.add_middleware(CORSMiddleware, allow_origins='*', allow_credentials=True, allow_methods=['*'],
+                               allow_headers=['*'])
+        app.mount("/static", StaticFiles(directory="/repo/data/web_gui/static"), name="static")
```
JsonSadler, line 36: Does "web_gui/static" comes from the frontend files? Sorry I have no clues on how these work together, just thought maybe it has not to be there if it is for the dashboard. Correct me if I am wrong

JsonSadler: maybe I am confused as the old name "web_gui" was used for front-end.

ui/web_gui.py
```diff
@@ -36,7 +42,6 @@ def create_fastapi_app(self):
                                allow_headers=['*'])
 
         app.mount("/panel/static", StaticFiles(directory="/srv/frontend/static"), name="panel")
-        app.mount("/static", StaticFiles(directory="/repo/data/web_gui/static"), name="static")
```
JsonSadler, line 39: Oh was this used for putting gstreamer files, etc. ? I think I now understand what you did, maybe it is better to change the name if it has to contain stream related files (playlist, video chunk, etc.)

api/processor_api.py
JsonSadler: Never mind, I think I figured it out and asked my question in "ui/web_gui.py" changes where you have removed this line from.

ui/web_gui.py
JsonSadler: Read several lines below and got it. Yeah, I think web_gui is not making sense as a name anymore, it's confusing.

api/processor_api.py
mats-claassen: This basically allows the access to the video streams. Renaming it sounds correct

ui/web_gui.py
mats-claassen: Right. I should make a more thorough search for web GUI mentions

mats-claassen: So Web_GUI now is the main class which serves the frontend files. It is decoupled from the video processing part. I replaced 'web_gui' with 'processor' in the path that serves the processed video stream.

frontend/src/Live.js
```diff
@@ -116,7 +116,7 @@ function Charts({cameras}) {
         if (!cameras) {
             return
         }
-        const url = `/static/data/${cameras[0]['id']}/objects_log/${new Date().toISOString().slice(0, 10)}.csv`;
+        const url = `${cameras[0]['storage_host']}/static/data/${cameras[0]['id']}/objects_log/${new Date().toISOString().slice(0, 10)}.csv`;
```
JsonSadler, line 119: Thanks @mats-claassen for the changes :) 

I was doing a little bit testing, ran API on 8000 and the GUI on 8003, it fails loading the videos and the plot, I tried debugging it seeing network requests in Firefox, seems the link to the csv file generated here is on port 8003 ( which is GUI's) not on 8000 (which is API's). I have no idea if this url string should refer to 8000 or this request from 8003 should internally be translated to 8000, but anyhow, the csv file is available on 8000 with the same url. 

Do you have any idea how to solve this?

mats-claassen: That url should point to the API, and not the GUI. Also `${cameras[0]['storage_host']}` should be something like `http://0.0.0.0:8000` in that case. Is that what you are seeing? Are you running everything on localhost?

JsonSadler: Then it is not the right url, it points to the GUI, and no, I am seeing that it is something like http://MYVPSIP:8003/static/data/default/objects_log/2020-06-25.csv , for example. while the API is running on the 8000. And yes both is running on the local host (my vps)

JsonSadler: @mats-claassen do you have any solution in mind? 
Let me know if you have tested and got different result

mats-claassen: So I have just tested it again. I run both frontend and API on the same host but access the web panel from another one. That is a small difference to your set up. Also I have changed two lines in the code (which are not pushed) to make this work: I changed the processor host to the correct IP in `config-frontend.ini` and I set `ENV DEV_ALLOW_ALL_ORIGINS=true` in `x86-openvino.Dockerfile` to allow access from that other host.
I could also try running everything on the same host.

mats-claassen: So, I tested doing everything from the same host and I still need to allow all origins in the processor node ( the `ENV DEV_ALLOW_ALL_ORIGINS=true` part). There is a note in the code with a warning regarding enabling this so I am hesitant to add it to all Dockerfile (jetson, Coral, Openvino, etc.). But I don't know if there is an alternative.

JsonSadler: Thanks for your help! Concurrently found several other issues with my setup which was the root of my confusion (frontend was also being copied in the processor dockerfiles - I am separating the API and Engine for the Processor too - I will fix this for all devices so don't worry).

The problem is now I get error 405 when the csv file is requested, and I guess this is the same thing you are telling with the `ENV DEV_ALLOW_ALL_ORIGINS=true`. I am testing and will tell you the result.

JsonSadler: @mats-claassen I could not get it to work with `ENV DEV_ALLOW_ALL_ORIGINS=true` in the dockerfile, would you please push your working state so I can be sure I am doing the whole process right?

JsonSadler: I still get 405 Error on retrieving the csv file (but the file is being downloaded if I copy and paste its link into my browser.
 [Screen shot](https://ibb.co/v39d6Df)

mats-claassen: The weird thing is that you get a 200 for the video playlist but can't access the csv. The only change I have is to add this line:
`ENV DEV_ALLOW_ALL_ORIGINS=true`
at line 82 of the Openvino Dockerfile. I will also test this on the x86 without Openvino and let you know but the change should be very similar

JsonSadler: Thanks a lot @mats-claassen for your help. It is working for your PR, and as my scenario was a little bit different, I could also get it to work with it ( I was trying to separate API and Engine in the processor, made a [PR here on another fork](https://github.com/emma-w-dev/smart-social-distancing/pull/1). )

@mhejrati  I think you can merge this PR in a temporary branch, so @emma-w-dev can make another PR on top of it, including further development of the new architecture with all the modules separated. Then there will be a huge PR after all to apply it to the master.

mats-claassen: Thanks @JsonSadler. Regarding the separation of API and Engine: how do you envision the communication between API and Engine for say, modify the configurations of the engine as in https://github.com/neuralet/neuralet/issues/25?
Which is the advantage of separating those two?

mats-claassen:
Closing as merged in #33

JsonSadler:
> Closing as merged in #33

Many thanks for your contribution :)

PR #24, opened 2020-06-26
Updated readme about coral's problem in building the frontend

JsonSadler:
A Self-explanatory PR :D


mhsekhavat:
Not only for coral, but for all edge devices :)

PR #31, opened 2020-07-15
New architecture developed

JsonSadler:
Hi, 

I have designed the new architecture in which processor's concept is separated into an API and a CORE, both running in a container which actually represents the "processor"'s concept (now runs only on x86 using OpenVino, other platforms should be added). Dashboard is separated based on branch of [separate_components](https://github.com/neuralet/smart-social-distancing/tree/separate_components).

Please see how the code works, and let me know where the architecture may improve.

Next, I am going to write support for all other platforms, and then I will try to merge #17 into my branch. 

Also thanks @emma-w-dev , @mats-claassen and @mhejrati  for their help. 

Maybe it is a good idea to make a branch for the new-architecture development and remove/merge extra branches into it so we can test all the changes afterwards.


JsonSadler:
This PR had some unnecessary changes, let me fix it.

PR #33, opened 2020-07-16
New architecture

emma-w-dev:
Hi,
We recently merged our efforts for making the new architecture working with @JsonSadler, changes in this (huge!) PR are listed bellow:
* It contains @mats-claassen work from branch `separate_components`, `run-frontend.Dockerfile` is now used to bring up the Dashboard's docker, which itself uses docker image for frontend, `neuralet/smart-social-distancing:latest-frontend` built with `frontend.Dockerfile`.
* The concept of Processor is now implemented in each dockerfile for each platform. Entrypoint is now a script which runs two parts of the Processor: processor_api and processor_core.
* There are some supported commands for the processor_core, in `share/commands.py` which are used for api and core to communicate. Core is listening on a queue for commands from the api, and puts back some feedback for the api on another queue. 
* Core's supported command are now:     
1- `PROCESS_VIDEO_CFG`: Runs in a thread, processes the video addressed in the config file. Does not locks the processor_core's functionality. 
2- `STOP_PROCESS_VIDEO`: Stops processing the video, if any is being processed at that time.
* API's supported paths are now:
1- `/process-video-cfg`: Sends command `PROCESS_VIDEO_CFG` to core and returns the response.
2- `/stop-process-video`: Sends command `STOP_PROCESS_VIDEO` to core and returns the response.
3- `/get-config`: It returns the config which is used by both api and core (it is the same so returns just a single config) in json format. 
4- `/set-config`: As the configuration file between api and core is the same configuration, it sets the given set of json configurations in the config, for both api and engine and reloads the configuration. Core's engine is also restarted so all methods and members (specially those which were initiated with the old config) can use the updated config (this will stop the processing of the video - if any). Comments on what to return is appreciated.
for this path, you can test it like entering this while the docker's are up and running: 
`curl -d '{"App": { "VideoPath" : "/repo/data/YOUR_VIDEO.mp4"} }' -H "Content-Type: application/json" -X POST http://localhost:8001/set-config` (of course you have to put your video under `data/` before) and then enter `http://127.0.0.1:8001/process-video-cfg` in your browser. You can see in your terminal running the docker that your video is being loaded and processed. You also can refresh your dashboard to see the output (TODO: residual files under data/web_gui/static/ may cause you to see previous streams and plots stored there! This needs to be issued separately)
* All configurations are implemented as classes in `api/config_keys.py` as `pydantic`'s `BaseModel` which is used to examine if the key and/or it's value are valid for setting or not (just parsed from json by FastApi, NO detailed sanity-check is done on the values). This fails the request to set some key which is not present in original setup.
* Test is written in `api/test_processor_api.py` using pytest (only for testing prosessor_api ), which is reading config-sample.ini file and send it as a post request to set-config and then get it through get-config api (get) call and assert if both are the same.


api/config_keys.py
```diff
@@ -0,0 +1,46 @@
...
+    Host: Optional[str]
+    Port: Optional[str]
+
+class Config(BaseModel):
+    App: Optional[APP]
+    Detector: Optional[DETECTOR]
+    PostProcessor: Optional[POSTPROCESSOR]
+    Logger: Optional[LOGGER]
+    API:  Optional[API_]
+    CORE: Optional[CORE_]
```
renzodgc, line 46: I don't think its ideal for the API to receive requests in PascalCase. snake_case seems to be what some would use nowadays with a python API.

With @mats-claassen we were working on https://github.com/neuralet/neuralet/issues/25 and by looking at FastAPI and Pydantic's API I implemented with snake_case (and some validation for video_path). We also found your commit about config_keys (that's why I used the same file name)

Here's the file if you want to look at it, https://github.com/xmartlabs/smart-social-distancing/blob/227d505ca643a252a7a4667ec91e60c633e32620/api/models/config_keys.py

My idea was to rebase with this after this PR was merged, adapt it to the newly proposed changes and open a new pr proposing these changes (along with a frontend implementation to edit video path)

What do you think?

api/config_keys.py
```diff
@@ -0,0 +1,46 @@
+from pydantic import BaseModel
+from typing import Optional
+
+
+class APP(BaseModel):
+    VideoPath: Optional[str]
+    Host: Optional[str]
+    Port: Optional[str]
```
renzodgc, line 8: Does it make sense to allow changing these as part of an endpoint request?

api/test_processor_api.py
```diff
@@ -0,0 +1,52 @@
...
+config_sample_json = {}
+
+for section in sections:
+    config_sample_json[section] = config_sample.get_section_dict(section)
+
+#@pytest.mark.order1
+def test_set_config():
+    response = client.post(
+    "/set-config",
+    json=config_sample_json,
```
renzodgc, line 34: Please indent this code  to reflect that's multiline inside the parenthesis (or change it to inline)

ui/web_gui.py
```diff
@@ -52,17 +57,36 @@ async def index():
...
-                    {'src': '/static/gstreamer/default-birdseye/playlist.m3u8', 'type': 'application/x-mpegURL',
+                    {'src': processor_host + '/static/gstreamer/default-birdseye/playlist.m3u8', 'type': 'application/x-mpegURL',
                      'birdseye': True},
                 ],
             }]
-
+        #@app.get("/static/")
+        #async def redirect():
+        #    response = RedirectResponse('127.0.0.1:8000/static/')
+        #    return response
```
renzodgc, line 74: Please remove code that's commented out

run-frontend.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+COPY ui/requirements.txt /ui/
+WORKDIR /ui
+
+RUN python3 -m pip install --upgrade pip setuptools==41.0.0 && pip install -r requirements.txt
+COPY ui/ /ui
+COPY libs/config_engine.py /ui/
+COPY config-frontend.ini /ui/
+
+#EXPOSE 8000
```
renzodgc, line 13: is this meant to be removed?

config-skeleton.ini
```diff
@@ -1,33 +1,35 @@
...
+QueuePort: 8010
+QueueAuthKey: shibalba
 
 [Detector]
-; Supported devices: Jetson , EdgeTPU, Dummy
-Device: EdgeTPU
-; Detector's Name can be either "mobilenet_ssd_v2", "pedestrian_ssdlite_mobilenet_v2" or "pedestrian_ssdlite_mobilenet_v2"
-; the first one is trained on COCO dataset and next two are trained on Oxford Town Center dataset to detect pedestrians
-Name: pedestrian_ssdlite_mobilenet_v2
-;ImageSize should be 3 numbers seperated by commas, no spaces: 300,300,3
```
renzodgc, line 14: I think we should leave these comments. I understand that the save function regenerates the whole `.ini`, do you know if this is doable with the `configparser` API? Or should we just make a config-skeleton.ini.example?

libs/processor_core.py
```diff
@@ -0,0 +1,93 @@
...
+            cmd_code = self._cmd_queue.get()
+            logger.info("command received: " + str(cmd_code))
+            
+            if cmd_code == Commands.RESTART:
+                # Do everything necessary ... make sure all threads in tasks are stopped 
+                if Commands.PROCESS_VIDEO_CFG in self._tasks.keys():
+                    logger.warning("currently processing a video, stopping ...")
+                    self._engine.stop_process_video()
+                
+                # TODO: Be sure you have done proper action before this so all threads are stopped
```
renzodgc, line 55: Is the idea to leave this TODO? Was it resolved by a commit?

run_processor_core.py
```diff
@@ -7,27 +7,21 @@
...
-        logger.error('video_path is not set in config file')
+def start_core(config):
+    from libs.processor_core import ProcessorCore
+    core = ProcessorCore(config)
     
-    logger.info("Engine terminated.")
-
+    logger.info("Core Started.")
+    core.start()
+    logger.info("Core Terminted.")
```
mats-claassen, line 16:
```suggestion
    logger.info("Core Terminated.")
```

libs/processor_core.py
```diff
@@ -0,0 +1,93 @@
...
+        self._tasks = {}
+       
+
+    def _setup_queues(self):
+        QueueManager.register('get_cmd_queue', callable=lambda: self._cmd_queue)
+        QueueManager.register('get_result_queue', callable=lambda: self._result_queue)
+        self._host = self.config.get_section_dict("CORE")["Host"]
+        self._queue_port = int(self.config.get_section_dict("CORE")["QueuePort"])
+        auth_key = self.config.get_section_dict("CORE")["QueueAuthKey"]
+        self._queue_manager = QueueManager(address=(self._host, self._queue_port), authkey=auth_key.encode('ascii'))
```
mats-claassen, line 30: What about using 'utf-8'?

amd64-usbtpu.Dockerfile
```diff
@@ -89,11 +91,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
     && apt-get autoremove -y
 
-ENTRYPOINT ["python3", "neuralet-distancing.py"]
-CMD ["--config", "config-skeleton.ini"]
-WORKDIR /repo
-EXPOSE 8000
+ENV DEV_ALLOW_ALL_ORIGINS=true
 
-COPY --from=neuralet/smart-social-distancing:latest-frontend /frontend/build /srv/frontend
+#COPY --from=neuralet/smart-social-distancing:latest-frontend /frontend/build /srv/frontend
```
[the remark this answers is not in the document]
emma-w-dev: yes, this line is moved to [run-frontend.Dockerfile](https://github.com/emma-w-dev/smart-social-distancing/blob/9d765c9a05bbe38dad1a21af98aa47a0f920c00a/run-frontend.Dockerfile#L3) and needs to be removed

api/config_keys.py
JsonSadler: Hm. This should be set in configuration which is used by the dashboard to connect to the api, so I think this is a wrong place for having it.

run-frontend.Dockerfile
JsonSadler: Yes, it's preferred to use no expose in the docker file as the port is configurable in the config file and in the instructions, we instruct to use the docker run with "-p" option

api/config_keys.py
emma-w-dev: This keys came from the config files as is in [configParser](https://docs.python.org/3/library/configparser.html) examples.
If you think it's better to follow the FastApi and Pydantic snake_case rules feel free to change it and good idea to change it after merge. thanks.

libs/processor_core.py
JsonSadler: It just informs future developers to remember what needs to be done here if they are adding further features. Now the only task maybe running is the task of processing the video which his taken care of

ui/web_gui.py
emma-w-dev: oo, this was for a test and is accidentally committed. thanks I'm going to remove it.

libs/processor_core.py
JsonSadler: Oh sure, as the whole idea of how to properly save and read the auth_key is not decided I just implemented a basic idea. Maybe it is better to just keep the binary or encoded or hashed or whatever - I am not experienced with authentication concerns.

config-skeleton.ini
emma-w-dev: You are right. There are solutions [here](https://stackoverflow.com/questions/6620637/writing-comments-to-files-with-configparser)

api/processor_api.py
```diff
@@ -0,0 +1,118 @@
...
+
+        logger.info("Connection established to Core's queue")
+        self._cmd_queue = self._queue_manager.get_cmd_queue()
+        self._result_queue = self._queue_manager.get_result_queue()
+
+    def create_fastapi_app(self):
+        # Create and return a fastapi instance
+        app = FastAPI()
+
+        if os.environ.get('DEV_ALLOW_ALL_ORIGINS', False):
```
mhejrati, line 62: @mats-claassen @mhsekhavat @emma-w-dev is this still relevant here? Assuming that the react GUI has been moved out of the processor alltogether.

Issue #25, opened 2020-06-30
XGD_Runtime Error and Loading Video problem


memina:
Hello, thank you to everyone who contributed to this project.
When I try it on Jetson nano, the video feels like loading continuously. It plays after 5 seconds, loads again, plays again for 5 seconds, then pauses. I refresh the page but the same problem continue. It also gives errors related to XGD_Runtime on console.

I created the docker image with the command below.
`docker build -f jetson-nano.Dockerfile -t "neuralet/smart-social-distancing:latest-jetson-nano" .`

Also I tried it on my PC, the result is same. What am i missing ? 

<img width="1323" alt="Screen Shot 2020-06-26 at 20 45 45" src="https://user-images.githubusercontent.com/12199743/86169699-b1883f80-bb22-11ea-88eb-b429edd4b5b1.png">

JsonSadler:
Hi,
Thanks for your feedback :)

It's weird. Can you try running the docker without sudo?

memina:
yes, but i can't run docker without sudo command.

JsonSadler:
> yes, but i can't run docker without sudo command.
Try:
`sudo usermod -aG docker $USER`
Then logout/login or reboot.

JsonSadler:
@memina Is this solved?

nabil-bukhari:
I am having the same issue without any display 

![Screenshot from 2020-07-30 18-48-52](https://user-images.githubusercontent.com/68958543/88944393-59656a00-d295-11ea-9487-d7f9a85012cf.png)

renzodgc:
Hey, we just got around checking these old issues, sorry about the delay.

The issue could be related to processing speed or the old frontend. Considering a few improvements have been done on the former and that we deprecated the old frontend for a new one on react, I presume this has been fixed. (Although the video streaming is still a bit rough around the edges)

Feel free to try to reproduce it if you are still interested and report back If performance is still an issue. Elsewise I'll close this issue in a week :)

PR #43, opened 2020-07-29
Modify configuration and parameters from web GUI

renzodgc:
This addresses [Neuralet Issue #25](https://github.com/neuralet/neuralet/issues/25).
- Updates the previously added endpoint to cope with the snake_case standard.
- Adds `video_path` validation in the API by trying to process the video through cv2 before attempting to update the config.
- Adds in the frontend a `/settings` page to update video config and point towards a processor's node API.
- The `/settings` page only shows up in the navbar if the frontend is run on dev mode (with npm).

image

I kept all other config options. However, they don't have validations, so misusing the endpoint may cause an invalid config file to be saved.


api/models/config_keys.py
```diff
@@ -15,7 +16,9 @@ class Config:
 
 
 class AppConfig(SnakeModel):
-    VideoPath: Optional[str] = Field(None, example='/repo/data/TownCentreXVID.avi')
+    VideoPath: Optional[str] = Field(None, example='/repo/data/gard1-4.mp4')
+    Host: Optional[str] = Field(None, example='0.0.0.0')
+    Port: Optional[str] = Field(None, example='8001')
```
mats-claassen, line 21: I think we should leave 8000 as example
```suggestion
    Port: Optional[str] = Field(None, example='8000')
```

mats-claassen: Actually, host and port are in ApiConfig so should be removed from here. or from ApiConfig

api/processor_api.py
```diff
@@ -99,38 +100,20 @@ async def stop_process_video():
...
 
         @app.put("/config")
-        async def update_config(config_request: ConfigRequest):
-            save_file = config_request.save_file
-            config_request = config_request.dict(exclude_unset=True, exclude_none=True)
-            config = config_request['config']
-            
+        async def update_config(config: Config):
+            config = config.dict(exclude_unset=True, exclude_none=True)
+            print(config)
```
mats-claassen, line 112: Please remove print

libs/processor_core.py
```diff
@@ -78,6 +78,7 @@ def _serve(self):
                     self._engine.stop()
                     del self._tasks[Commands.PROCESS_VIDEO_CFG]
                     logger.info("processing stopped")
+                    self.config.reload()
```
mats-claassen, line 81: Why is this called in STOP? Wouldn't it be better to do this right before we create an EngineConfig in `if cmd_code == Commands.PROCESS_VIDEO_CFG`?

mhejrati:
@mats-claassen have you tested this already or does it need testing on different platforms before merging?

mats-claassen:
Has been tested on x86-Openvino and x86

mats-claassen:
Tested also on Jetson Nano. I think this is ready to be merged

PR #52, opened 2020-08-11
add calibrated method for distance calculation

alpha-carinae29:


libs/distancing.py
```diff
@@ -363,8 +375,9 @@ def calculate_box_distances(self, nn_out):
 
         """
         This function calculates a distance matrix for detected bounding boxes.
-        Two methods are implemented to calculate the distances, first one estimates distance of center points of the
-        boxes and second one uses minimum distance of each of 4 points of bounding boxes.
+        Three methods are implemented to calculate the distances,first one estimates distance with a calibration matrix
```
mats-claassen, line 378:
```suggestion
        Three methods are implemented to calculate the distances, the first one estimates distance with a calibration matrix
```

alpha-carinae29: Done, thanks :)

mats-claassen (approved):

libs/distancing.py
```diff
@@ -30,6 +31,18 @@ def __init__(self, config):
...
         self.birds_eye_resolution = (200, 300)
+        if self.dist_method == "CalibratedDistance":
+            try:
+                calibration_file = self.config.get_section_dict("App")["CalibrationFile"]
+            except KeyError:
+                raise ValueError(
+                    "The 'CalibrationFile' should be specified in config file in case of using 'CalibratedDistance' method")
+            try: 
+                with open(calibration_file, "r") as file:
+                    self.h_inv = file.readlines()[0].split(" ")[1:]
```
mats-claassen, line 42: Hey @alpha-carinae29, which is the format you are expecting for this file? I understand that it should contain the homography matrix. But what is the first value that you are discarting here?

alpha-carinae29: Hey @mats-claassen, this is the inverse homography matrix file of the soft-bio dataset:
`h_inv: 1.40165283e-02 1.89412544e-04 -7.09122683e+00 8.32773572e-04 -1.84875733e-02 7.68117031e+00 2.87659642e-04 1.18484948e-03 3.37206106e-01`
where each number is one of the elements of inverse homograpgy matrix with this format:
`h_inv: h_11 h_12 h_13 h_21 h_22 h_23 h_31 h_32 h_33` where `h_ij` is the element of the i-th row and j_th columns of the matrix.

alpha-carinae29: I had included this sample matrix for soft-bio in my fork, it seems I forgot to commit it.

alpha-carinae29: I just found out that the `data/` directory is in `gitignore` so my default homography file had not been pushed.

renzodgc: Hello @alpha-carinae29 ! I'm working with @mats-claassen on integrating this feature to the processor's API, and I have a question regarding this.

When you mean the inverse homography matrix `(h_inv)` it means I should apply something like `numpy.linalg.inv()` to the result of `h, _ = cv2.findHomography(pts_source), pts_destination))` in order to get the inverse of the homography matrix `(h)` is that correct? :)

alpha-carinae29: Hey @renzodgc, Yeah you are totally right, the `cv2.findHomography(pts_source), pts_destination))` returns a matrix that can transform the real world coordinates to the image coordinates, However we want exactly the opposite way, we should transform the image coordinate to the world coordinate to calculate distances correctly. So we should use the inverse matrix.
If this explanation has ambiguity, please don't hesitate to ask more questions. Thanks.

renzodgc: Perfectly clear! Thank you!

PR #63, opened 2020-10-05
Fix: Send empty reports when there is no data for a camera instead of 404

renzodgc:
Previously the reports endpoints checked whether reports for a camera were generated or not.

Now a camera without reports is considered the same as having no detections/violations and returns an empty report for said dates expected.

### DIFF
--- a/config-x86.ini
+++ b/config-x86.ini
@@ -11,44 +11,33 @@ QueueAuthKey = shibalba
 Resolution = 640,480
 Encoder = videoconvert ! video/x-raw,format=I420 ! x264enc speed-preset=ultrafast
 MaxProcesses = 1
-; Screenshot time is measured in minutes (if period <= 0 then no screenshots are uploaded)
 ScreenshotPeriod = 0
 ScreenshotS3Bucket = my-screenshot-bucket
-; WIP https://github.com/neuralet/neuralet/issues/91
-;Encoder: videoconvert ! vaapih264enc
 DashboardURL = http://0.0.0.0:8000
 ScreenshotsDirectory = /repo/data/processor/static/screenshots
 CalibrationFile = /repo/data/homography_matrix/h_inverse.txt
 
 [Source_0]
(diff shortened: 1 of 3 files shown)


libs/utils/reports.py
```diff
@@ -207,17 +215,23 @@ def peak_hour_violations(self, camera_id):
...
         log_dir = os.getenv('LogDirectory')
         file_path = os.path.join(log_dir, camera_id, "objects_log", "report.csv")
+        if not os.path.exists(file_path):
+            return 0.0
         df = pd.read_csv(file_path).drop(['Number'], axis=1)
         return round(df['ViolatingObjects'].mean(), 1)
 
     def camera_with_most_violations(self):
         log_dir = os.getenv('LogDirectory')
         cameras = os.listdir(log_dir)
```
mats-claassen, line 225: I don't think this is the correct way to check for a list of cameras. You should get them from the config file. There could be other files and folders in there, not just `.keep`

renzodgc: The issue is that this file doesn't have access to the configuration (since we can't deliver the object to the reports subapi). Maybe we could add the .keep to the docker ignore?

renzodgc: Hm no, that doesn't work because data is a mounted volume

config-x86.ini
```diff
@@ -11,44 +11,33 @@ QueueAuthKey = shibalba
 Resolution = 640,480
 Encoder = videoconvert ! video/x-raw,format=I420 ! x264enc speed-preset=ultrafast
 MaxProcesses = 1
-; Screenshot time is measured in minutes (if period <= 0 then no screenshots are uploaded)
```
mats-claassen, line 14: Please leave the comments

mats-claassen: I think you should discard all changes to this file

mats-claassen (approved):

PR #70, opened 2020-11-06
Occupancy Notifications

renzodgc:
Added a process with threads per each area that notifies for occupancy


libs/area_reporting.py
```diff
@@ -0,0 +1,72 @@
...
+        self.mail_service = MailService(config)
+        self.slack_service = SlackService(config)
+
+    def process_area(self):
+        # Sleep for a while so cameras start processing
+        time.sleep(30)
+
+        self.processing_alerts = True
+        logger.info(f'Enabled processing alerts for - {self.area_id}: {self.area_name} with {len(self.cameras)} cameras')
+        while self.cameras and self.processing_alerts:
```
pgrill, line 43: Why we need the `self.cameras` condition in the while true?
Taking into accout that all the areas should have unless a camera this condition will always be `True`

libs/area_reporting.py
```diff
@@ -0,0 +1,72 @@
...
+            if not all(list(map(os.path.isfile, camera_file_paths))):
+                # Wait before csv for this day are created
+                logger.info(f'Area reporting on - {self.area_id}: {self.area_name} is waiting for reports to be created')
+                time.sleep(5)
+
+            occupancy = 0
+            for camera in self.cameras:
+                with open(os.path.join(camera['file_path'], str(date.today()) + ".csv"), 'r') as log:
+                    last_log = deque(csv.DictReader(log), 1)[0]
+                    occupancy += int(last_log['DetectedObjects'])
```
pgrill, line 54: I think that we should also check the `last_log` time.
If for some reason a camera stops to register the information in the csv log, the occupancy variable will summarize objects from different hours.

libs/area_reporting.py
```diff
@@ -0,0 +1,72 @@
...
+        self.cameras = [camera for camera in self.config.get_video_sources() if camera['id'] in self.area['cameras']]
+        for camera in self.cameras:
+            camera['file_path'] = os.path.join(self.log_dir, camera['id'], "objects_log")
+            camera['last_processed_time'] = time.time()
+
+        self.mail_service = MailService(config)
+        self.slack_service = SlackService(config)
+
+    def process_area(self):
+        # Sleep for a while so cameras start processing
```
pgrill, line 38: What happens if the area doesn't have the notification enabled?.
The thread will process the CSVs without any purpose.

Maybe we can add the sentence 
```
if not (self.should_send_email_notifications or self.should_send_slack_notifications):
    return
```

config-coral.ini
```diff
@@ -9,6 +9,8 @@ DashboardURL = http://0.0.0.0:8000
 ScreenshotsDirectory = /repo/data/processor/static/screenshots
 EnableSlackNotifications = no
 SlackChannel = lanthorn-notifications
+; OccupancyAlertsTimeout time is measured in seconds (if timeout <= 0 then no occupancy alerts are triggered)
```
mats-claassen, line 12: At first reading this it is not immediately clear to me what this parameter does. I guess that can be solved by documenting the config file parameters in the future. Or maybe we could change the name. What does Timeout refer to? Normally it would be something where you make a request and wait for Timeout seconds before cancelling the request.

If the idea of this is similar to the Logger -> TimeInterval then a similar name would make sense I think. Something like OccupancyAlertsMinInterval ?

libs/area_reporting.py
```diff
@@ -0,0 +1,72 @@
...
+class AreaReporting:
+
+    def __init__(self, config, area):
+        self.processing_alerts = False
+        self.config = config
+        self.area = area
+
+        self.occupancy_sleep_time_interval = float(self.config.get_section_dict("App")["OccupancyAlertsTimeout"])
+        self.log_dir = self.config.get_section_dict("Logger")["LogDirectory"]
+        self.idle_time = float(self.config.get_section_dict('Logger')['TimeInterval']) + 0.1
```
mats-claassen, line 22: If the idea is to process this for every frame processed from the camera then here you will be missing a few frames every minute. I think the ideal would be to sleep For TimeInterval seconds. Or maybe we don't want to check this as often. Maybe it is enough to check it every few seconds?

libs/area_threading.py
```diff
@@ -0,0 +1,42 @@
...
+from libs.distancing import Distancing as CvEngine
+from libs.area_reporting import AreaReporting as AreaEngine
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def run_area_processing(config, pipe, areas):
+    pid = os.getpid()
+    logger.info(f"[{pid}] taking on {len(areas)} cameras")
```
mats-claassen, line 13: This looks like copied from the other file. Could possibly be removed. Or the text updated

libs/area_reporting.py
renzodgc: Well technically one could edit the config.ini manually, I think it checking this "just in case" is not a bad idea. Although probably some other things break without this.

I'd still leave it until we validate the config files before starting up the processor

libs/area_reporting.py
renzodgc: Yeah, I thought about doing this but I wasn't sure if it was worth it. I fought it by adding the timer so the cameras start processing before this thread, but I don't know exactly how we should cope with a camera that stops logging data (Whether should we ignore it for the occupancy of the area or do something as take the median of the last 10 readings)

libs/area_reporting.py
renzodgc: I'll add this condition to https://github.com/neuralet/smart-social-distancing/pull/70/files#diff-2c0d82fa7dc706795dbef4565a49715e6de108f3c09c7c12b4f30d165b62ad45R138 so the thread isn't created in that case

config-coral.ini
renzodgc: Yeah interval is also not the word but it does sound better than timeout

libs/area_reporting.py
pgrill: Ok, I that case maybe we can add the condition
```
if not self.cameras:
   return
```
It sounds weird to me a `while` condition that will not change in the middle of the execution.
What do you think?

renzodgc: I ended up adding it so that it doesn't create the thread if the area has no cameras

libs/area_reporting.py
renzodgc: Yeah, I think you're right. I thought about putting a small delay so that if a processor takes longer then it would be included, but this would cause the processing to be even more unpredictable

libs/area_threading.py
```diff
@@ -0,0 +1,42 @@
+import os
+from shutil import rmtree
```
pgrill, line 2: I think that the followinf importst are not used
```
 from shutil import rmtree
 from libs.distancing import Distancing as CvEngine
```

config-coral.ini
```diff
@@ -9,6 +9,8 @@ DashboardURL = http://0.0.0.0:8000
 ScreenshotsDirectory = /repo/data/processor/static/screenshots
 EnableSlackNotifications = no
 SlackChannel = lanthorn-notifications
+; OccupancyAlertsMinInterval time is measured in seconds (if timeout <= 0 then no occupancy alerts are triggered)
```
pgrill, line 12: I didn't find the implementation of `(if timeout <= 0 then no occupancy alerts are triggered)`.
I think that we should implement the validation or remove the comment.

pgrill (approved):

mats-claassen (approved):

PR #74, opened 2020-11-11
Get raw data endpoints

pgrill:
- Add the `/reports/{camera_id}/daily_data` endpoint to retrieve the daily CSV file without any analysis
- Include the webhooks support in loggers
- Refactor the loggers removing unused code.


libs/loggers/web_hook_logger.py
```diff
@@ -0,0 +1,29 @@
...
+    def log_objects(self, objects, violating_objects, violating_objects_index_list, violating_objects_count,
+                    detected_objects_cout, environment_score, time_stamp, version):
+        request_data = {
+            'Version': version,
+            'Timestamp': time_stamp,
+            'DetectedObjects': detected_objects_cout,
+            'ViolatingObjects': violating_objects_count,
+            'EnvironmentScore': environment_score,
+            'Detections': str(objects),
+            'ViolationsIndexes': str(violating_objects_index_list)
```
renzodgc, line 24: Does it make sense to go for PascalCase here?

libs/loggers/csv_logger.py
```diff
@@ -55,68 +19,33 @@ def __init__(self, config, camera_id):
...
-                writer.writerows(object_dict)
-
-    def log_distances(self, distances, frame_number, file_path):
-        """Write violated incident's information of a frame into the object log file.
-
-        Each row of the distances log file consist of a violation information such as object (person) ids,
-        distance between these two object and frame number.
-
         Args:
+            objects_list: A list of dictionary where each dictionary stores information of an object (person) in a frame.
```
mats-claassen, line 32: I guess this parameter list should be updated

libs/loggers/loggers.py
```diff
@@ -46,14 +54,42 @@ def update(self, objects_list, distances):
...
+                violating_objects,
+                violating_objects_index_list,
+                violating_objects_count,
+                detected_objects_count,
+                environment_score,
+                current_time,
+                version=LOG_FORMAT_VERSION
+            )
+            if self.web_hook_logger:
+                self.web_hook_logger.log_objects(
```
mats-claassen, line 82: As these two loggers have a log_objects function  with the same signature, I am tempted to do something like this:

```python
for logger in [logger1, logger2]:
    logger.log_objects(...)
```

That would be nice if you plan to support multiple loggers, otherwise might not be that helpful.

What do you think?

libs/loggers/web_hook_logger.py
pgrill: I just copied the CSV file header. But you are right, PascalCase doesn't make any sense here.
I changed it to `snake_case` to follow the other endpoints format.

libs/loggers/csv_logger.py
pgrill: Yes, it was the documentation of the previous function.
Taking into accout that probably we will refactor the loggers I removed the comment for now.
We can document everything after the refactor.

libs/loggers/loggers.py
pgrill: I think that this is a good idea.

I changed the logger implementation structure following the pattern you suggested.

libs/loggers/csv_logger.py
```diff
@@ -55,68 +21,23 @@ def __init__(self, config, camera_id):
...
-        with open(file_path, "a", newline="") as csvfile:
-            field_names = ["frame_number", "object_0", "object_1", "distance"]
-            writer = csv.DictWriter(csvfile, fieldnames=field_names)
-            writer.writerows([{"frame_number": frame_number,
-                               "object_0": indices[0],
-                               "object_1": indices[1],
-                               "distance": distances[indices[0], indices[1]]} for indices in violating_objects])
+
+            writer.writerow(
+                {'Version': version, 'Timestamp': time_stamp, 'DetectedObjects': detected_objects_cout,
```
mats-claassen, line 41: You are using single quotes here. Do we want to use " everywhere?

config-jetson.ini
```diff
@@ -83,3 +83,4 @@ TimeInterval = 0.5
 LogDirectory = /repo/data/processor/static/data
 EnableReports = no
 HeatmapResolution = 150,150
+WebHooksEndpoint=
```
mats-claassen, line 86: This has to be documented in one of these PRs, depending which gets merged first

libs/loggers/csv_logger.py
pgrill: I changed the quotes from `'` to `"`. I also apply the change in the `web_hook_logger.py `

config-jetson.ini
pgrill: I included the `WebHooksEndpoint` documentantion in the README.

mats-claassen (approved):

renzodgc (approved):

PR #77, opened 2020-11-13
Facemask improvements.

pgrill:
Adding facemask label in boxes.
Including the minscore threshold for mask detection in the classifier.

### DIFF
--- a/create_reports.py
+++ b/create_reports.py
@@ -71,7 +71,8 @@ def create_daily_report(config):
                 hour = datetime.strptime(row['Timestamp'], "%Y-%m-%d %H:%M:%S").hour
                 detections = ast.literal_eval(row['Detections'])
                 faces = [d["face_label"] for d in detections if "face_label" in d]
-                summary[hour] += (1, int(row['DetectedObjects']), int(row['ViolatingObjects']), len(faces), int(sum(faces)))
+                masks = [f for f in faces if f == 0]
+                summary[hour] += (1, int(row['DetectedObjects']), int(row['ViolatingObjects']), len(faces), len(masks))
 
         with open(daily_csv, "w", newline='') as csvfile:
             headers = ["Number", "DetectedObjects", "ViolatingObjects", "DetectedFaces", "UsingFacemask"]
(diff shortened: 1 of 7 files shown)


create_reports.py
```diff
@@ -71,7 +71,8 @@ def create_daily_report(config):
                 hour = datetime.strptime(row['Timestamp'], "%Y-%m-%d %H:%M:%S").hour
                 detections = ast.literal_eval(row['Detections'])
                 faces = [d["face_label"] for d in detections if "face_label" in d]
-                summary[hour] += (1, int(row['DetectedObjects']), int(row['ViolatingObjects']), len(faces), int(sum(faces)))
+                masks = [f for f in faces if f == 0]
```
mats-claassen, line 74: So `d["face_label"] == 0` when the person is using a facemask? That sounds weird

libs/distancing.py
```diff
@@ -68,6 +68,9 @@ def __init__(self, config, source, live_feed_enabled=True):
         if not os.path.exists(self.screenshot_path):
             os.makedirs(self.screenshot_path)
 
+        if "Classifier" in self.config.get_sections():
+            self.face_threshold = float(self.config.get_section_dict("Classifier").get("MinScore", 0.75))
```
mats-claassen, line 72: Is there a special reason to use 0.75 as default score threshold, and not 0.5 for example?

create_reports.py
pgrill: The facemask classifier returns `class_ids`, not boolean values. This can be read in the docstring of the function.
```
"""
        Inference function sets input tensor to input image and gets the output.
        The interpreter instance provides corresponding class id output which is used for creating result
        Args:
            resized_rgb_images: Array of images with shape (no_images, img_height, img_width, channels)
        Returns:
            result: List of class id for each input image. ex: [0, 0, 1, 1, 0]
            scores: The classification confidence for each class. ex: [.99, .75, .80, 1.0]
        """
```
In the tests that I did, the value 0 is attached to the class `with mask`.

libs/distancing.py
pgrill: I put the value 0.75 to improve the accuracy in the mask detection reducing the number of false positives.

libs/utils/visualization_utils.py
```diff
@@ -358,6 +362,9 @@ def visualize_boxes_and_labels_on_image_array(
                     box_to_color_map[box] = STANDARD_COLORS[
                         classes[i] % len(STANDARD_COLORS)
                         ]
+                if face_labels:
+                    facemask_label = f"Facemask: {face_index[face_labels[i]]}"
```
renzodgc, line 366: the label is just text?

renzodgc (commented):
Could you attach a screenshot on how the facemask tagging looks in the video feed?

libs/utils/visualization_utils.py
pgrill: Yes, it's new a label bellow to the current `Pedestrian ..` showing the mask usage.
Screenshot from 2020-11-16 10-13-29

renzodgc (approved):

mats-claassen (approved):

mrn-mln (approved):

PR #82, opened 2020-11-18
Fault tolerance

pgrill:


libs/area_reporting.py
```diff
@@ -31,7 +31,8 @@ def __init__(self, config, area):
             camera['last_processed_time'] = time.time()
 
         self.mail_service = MailService(config)
```
renzodgc, line 33: I know this wasn't edited in this change, but presumably an `if self.should_send_email_notifications` should be here, no? :sweat_smile:

renzodgc:
```suggestion
        if self.should_send_email_notifications:
            self.mail_service = MailService(config)
```

libs/area_threading.py
```diff
@@ -33,6 +34,16 @@ def __init__(self, config, area):
...
+            try:
+                self.engine.process_area()
+            except Exception as e:
+                logging.error(e, exc_info=True)
+                # Sleep the thread for 5 seconds and try to process the area again
+                time.sleep(5)
+                logging.info(f"Exception processing area {self.area['name']}")
+                logging.info("Restarting the area processing")
+
         self.engine.process_area()
```
renzodgc, line 47: Should this be deleted?

pgrill: Maybe we can delete the `time.sleep(5)`. 
I put it because when the `Exception` is fatal, you get the log flooded by the Exception trace. However, the `sleep` doesn't fix that issue. I will change it implementing a max retry logic.

amd64-usbtpu.Dockerfile
```diff
@@ -94,6 +95,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 ENV DEV_ALLOW_ALL_ORIGINS=true
 ENV AWS_SHARED_CREDENTIALS_FILE=/repo/.aws/credentials
 ENV AWS_CONFIG_FILE=/repo/.aws/config
+ENV CONFIG_FILE=config-coral.ini
```
mats-claassen, line 98: If you want to use another config file you would now have to rebuild the steps behind this so I would prefer to have it at the end

supervisord.conf
```diff
@@ -0,0 +1,42 @@
...
+[rpcinterface:supervisor]
+supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
+[supervisorctl]
+serverurl=unix:///var/run/supervisor.sock
+
+[program:api]
+command=python3 run_processor_api.py --config %(ENV_CONFIG_FILE)s
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
```
mats-claassen, line 16: maxbytes = 0 means unlimited? What if we put a limit here, say 200Mb for each log? If this is going to be running possibly for years then that would be good

pgrill: The `stdout_logfile_maxbytes` parameter defines the number of bytes required to rotate the log file (http://supervisord.org/logging.html#activity-log-rotation). The value 0 means `the logfile is never rotated `. 
We are using `/dev/stdout` as logfile, so the rotation is not required.

amd64-usbtpu.Dockerfile
pgrill: You are right. I changed it.

libs/area_threading.py
pgrill: I added the `MaxThreadRestarts` parameter in the config file.
I also removed the line `self.engine.process_area() `.

renzodgc (approved):

libs/engine_threading.py
```diff
@@ -44,15 +44,20 @@ def __init__(self, config, source, live_feed_enabled=True):
...
         self.engine = CvEngine(self.config, self.source['section'], self.live_feed_enabled)
+        restarts = 0
+        max_restarts = int(self.config.get_section_dict("App")["MaxThreadRestarts"])
         while True:
             try:
                 self.engine.process_video(self.source['url'])
             except Exception as e:
                 logging.error(e, exc_info=True)
+                logging.info(f"Exception processing video for source {self.source['name']}")
+                if restarts == max_restarts:
```
mats-claassen, line 55: `restarts` never gets reset so this is counting restarts no matter how much time passed between one and the next (could be a week for example). Not sure we need to go so deep and improve that but we could (save time of last restart and if it is greater than 1/2/3 minutes then reset count to 0)

pgrill: Yes, I wasn't sure how much complexity to include in the logic. But make sense to take into account the time variable. I will include it.

pgrill: Done

mats-claassen (approved):

PR #84, opened 2020-11-19
Processor Basic Info Endpoint. HasBeenConfigured Config

renzodgc:
- Added an endpoint `GET /config/info` that returns some basic info and processor's version.
- Added a new config under `App`, `HasBeenConfigured` that is False by default and updates to  True if the config was updated.
- Modified README to let users know that if they manually change the setup they should also change this config.
- Added a constants.py file
- Added a .env and .env.example (moved email there)


api/config.py
```diff
@@ -33,6 +49,18 @@ def map_config(config, options):
     }
 
 
+def processor_info(config):
+    has_been_configured = bool(config["App"]["HasBeenConfigured"])
+    device = config["Detector"]["Device"]
+    if config["Detector"]["Name"] == "openvino":
+        device += "-openvino"
+    return {
+        "version": "0.2.0",
```
pgrill, line 58: Can we move the version number to an independent file?
It should be in a place easy to find and edit each time we deploy a new one.

pgrill (approved):

README.md
```diff
@@ -212,12 +212,16 @@ You can read and modify the configurations in `config-*.ini` files, accordingly:
...
+Please note that if you modify these values you should also set `[App]` `HasBeenConfigured` to `"True"`.
+This allows for a client to recognize if this processor was previously configured.
+
 You can also modify some of them using the [UI](https://beta.lanthorn.ai). 
 If you choose this option, make sure to mount the config file as a volume to keep the changes after any restart of the container.
 
 All the configurations are grouped in *sections* and some of them can vary depending on the chosen device.
 
 - `[App]`
+  - `HasBeenConfigured`: A boolean parameter that state if the *config.ini* was changed or not.
```
mats-claassen, line 224:
```suggestion
  - `HasBeenConfigured`: A boolean parameter that states whether the *config.ini* was set up or not.
```

constants.py
```diff
@@ -0,0 +1 @@
+PROCESSOR_VERSION = "0.2.0"
```
mats-claassen, line 1: I think this should be set when a release is made, but let's suppose we will create it soon

mats-claassen (approved):

constants.py
mats-claassen: I mean that technically we are still 0.1.0

renzodgc: That's true, we could include the changelog in this pull request

renzodgc, line 1:
```suggestion
PROCESSOR_VERSION = "0.1.0"
```

PR #86, opened 2020-11-20
Enable slack notifications per entity

renzodgc:
Move Slack Configuration from a global config to each area/camera


api/models/config_keys.py
```diff
@@ -24,6 +24,7 @@ class NotificationConfig(SnakeModel):
     violationThreshold: Optional[int] = Field(0, example=100)
     notifyEveryMinutes: Optional[int] = Field(0, example=15)
     emails: Optional[str] = Field("", example='john@email.com,doe@email.com')
+    enableSlackNotifications: Optional[str] = Field(False, example=False)
```
pgrill, line 27: Maybe we can declare it as a bool field.

renzodgc: I think it needs to be a string to be written in the config file

pgrill: :thinking: But the daily report is declared as `bool` and it works.
If we declare it as a string, the field definition should be `Field("False", example="False")`

mats-claassen (approved):

api/models/config_keys.py
renzodgc: You were right :), I got mixed up

pgrill (approved):

Issue #88, opened 2020-11-21
Master is broken for x86-gpu


JsonSadler:
Hi, 

Master does not work well for x86-gpu, seems prints from the `Detector` class are not showing (the first thing that made me think it is not working), but I can see some outputs in Lanthorn UI, which stops after a few frames and fps becomes zero.

The weird thing is that the lines:
`INFO:libs.distancing:processed frame 1 for /repo/data/softbio_vid.mp4
INFO:openpifpaf.decoder.generator.cifcaf:3 annotations: [13, 10, 9]
...
`
are showing for x86, but not for x86-gpu (while it processes some frames initially). What has been changed?

mrn-mln:
Hi @JsonSadler, the x86-gpu config file must have tracker parameters which I added, however, I have the same issue.

JsonSadler:
Hi @mrn-mln , thanks. Also needed `WebHooksEndpoint=` under `[Logger]`, which you fixed and thanks for that, BUT!

I tried with the config file in #89 , but it is still not working on master, but initially works with this commit: 09559b1800f84f9ce0cc49a805cce1bfe2802ff7

I tried to do a git bisect - of course needed to manually replace the config file from #89 - and found that this is the bad commit: bf82aadeeff2bac1eb6e35a3e964fa5a4b4b010d
where for the first time x86-gpu does not work. Seems it exits for some reason without telling why, printing:
`INFO:libs.engine_threading:[173] taking on 1 cameras
ok video is going to be processed
2020-11-21 06:02:40,154 INFO exited: startup (exit status 0; expected)`
and then it stucks  there ... 

I am going to debug this, any help will be appreciated @pgrill @renzodgc

mats-claassen:
Hi @JsonSadler and @mrn-mln, I just tested master using `x86-gpu.Dockerfile` and `config-x86-gpu.ini` after your fixes in #89 and it works for me, both running OpenPifPaf and mobilenet. With my GPU, I can see the live feed on the Lanthorn web when I run mobilenet but OpenPifPaf doesn't seem to generate enough FPS. The only change that I have locally is that I enabled SSL.

Did you find any clue to your issue with your debugging?

mrn-mln:
I didn't enable SSL.
Thanks @mats-claassen

renzodgc:
@JsonSadler did enabling SSL solved that for you? (I presume yes since this issue is quite old :sweat_smile:)

[...]

PR #89, opened 2020-11-21
added tracker parameters to gpu config

mrn-mln:
#88


JsonSadler (changes requested):
seems also needs `WebHooksEndpoint=` under `[Logger]`

mhejrati (commented):
LGTM

PR #94, opened 2020-11-26
Add export_all endpoint

mats-claassen:
This endpoint allows downloading all raw csv data for all currently configured cameras

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -1,5 +1,7 @@
 import logging
 import humps
+import os
+import shutil
 
 from fastapi import status
 from fastapi.encoders import jsonable_encoder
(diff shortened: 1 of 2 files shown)


api/utils.py
```diff
@@ -77,3 +79,15 @@ def reestructure_areas(config_dict):
             config_dict[f"Area_{index}"] = config_dict[area_name]
             config_dict.pop(area_name)
     return config_dict
+
+
+def clean_up_file(filename):
```
renzodgc, line 84: How long does this take to execute? Can the file be deleted before its download is finished?

pgrill (approved):

renzodgc (approved):

PR #98, opened 2020-12-04
Extend config api

pgrill:
Allow retrieving and update all the sections in the configuration file using the API


mats-claassen (approved):

api/routers/periodic_tasks.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+@periodic_tasks_router.post("", response_model=PeriodicTaskDTO,
+                            status_code=status.HTTP_201_CREATED, response_model_exclude_none=True)
+async def create_periodic_task(new_periodic_task: PeriodicTaskDTO, reboot_processor: Optional[bool] = True):
+    """
+    Adds a periodic task.
+    """
+    if new_periodic_task.name != "reports":
+        raise ValidationError(f"Not supported periodic task named: {new_periodic_task.name}")
```
renzodgc, line 52: Same than my last comment, these controllers are pretty complex, maybe it makes sense to move what logic we can to the models

api/routers/area_loggers.py
```diff
@@ -0,0 +1,121 @@
...
+    config = extract_config(config_type="area_loggers")
+    return [map_section_from_config(x, config) for x in config.keys()]
+
+
+def get_area_loggers_model(logger):
+    if logger.name == "file_system_logger":
+        return FileSystemLoggerDTO
+    else:
+        raise ValueError(f"Not supported logger named: {logger.name}")
+
```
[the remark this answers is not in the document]
pgrill: For now, we have a single area logger. However, in the future, we will have multiple of them (an S3 logger for example) with different parameters. For that reason, we need different models and choose the right one by the `Name`.

api/routers/periodic_tasks.py
pgrill: Are you talking about moving the name validation into the model or the whole logic?

api/models/source_logger.py
```diff
@@ -34,3 +40,16 @@ class WebHookLogger(OptionalSectionConfig):
 
 class SourceLoggerListDTO(SnakeModel):
     sourcesLoggers: List[SourceLoggerDTO]
+
+
+def validate_logger(logger: SourceLoggerDTO):
```
renzodgc, line 45: weren't these also supposed to return the class in order to create the object?

renzodgc: Or creating the object is just for validation purposes and  the config is edited regardless of the created object?

pgrill: Creating the object is just for validation purposes (for example, a mandatory field is not sent.)

renzodgc (approved):

PR #99, opened 2020-12-11
New metrics dashboard

pgrill:
Metrics improvements:

    - Social distancing infringements were split into categories (Critical, Low, etc)
    - Face mask metrics were also improved
    - Occupancy metrics were added
    -  The reports' endpoints were replaced by the metrics' ones, adding now live data.


libs/metrics/social_distancing.py
```diff
@@ -0,0 +1,187 @@
...
+    @classmethod
+    def process_distance_violation_for_object(cls, distance_violations: List[dict]) -> Tuple[int, int]:
+        """
+        Receives a list with the "social distancing detections" (for a single person) and returns a
+        tuple with the summary of detections and violations (grouped by severity). Consecutive detections in
+        the same state are grouped and returned as a single one. Detections lower than the constant
+        PROCESSING_COUNT_THRESHOLD are ignored.
+
+        The infrigement categories are :
+            - Low: Between 10 seconds 30 seconds
```
mats-claassen, line 58: I know these numbers can be changed but where did we get them from? If the idea is to avoid COVID then we should base it on research and less than 5 or maybe even 15 minutes should be low. Critical I would say could be something like 2 hours. But we should validate those numbers

libs/metrics/social_distancing.py
```diff
@@ -0,0 +1,187 @@
...
+                    processing_count = 0
+                processing_count += 1
+                if current_status != processing_status and processing_count >= cls.processing_count_threshold:
+                    # Object was enouth time in the same state, change it
+                    if current_status is not None:
+                        # Append the previous status in the detections list
+                        seconds_in_status = (dist_violation["time"] - processing_time).seconds
+                        detections.append({"status": status, "seconds": seconds_in_status})
+                    current_status = processing_status
+                    processing_time = dist_violation["time"]
```
mats-claassen, line 89: I think this line should be in the if

libs/metrics/social_distancing.py
```diff
@@ -0,0 +1,187 @@
...
+            # Append the latest status
+            seconds_in_status = (distance_violations[-1]["time"] - processing_time).seconds
```
mats-claassen, line 91: This is run hourly so we cannot detection infringements of more than 60 minutes. I think for a first version it is ok but in the future we could store info about the "open infringements" from one hour and continue processing them in the next hour

libs/metrics/social_distancing.py
pgrill: I added a comment mentioning that the values only have demo purposes a needs to be checked with the OMS recommendations. If we use the real values, most of the video demo that we have will only generate low infractions.

libs/metrics/social_distancing.py
pgrill: Yes, I agree. Maybe we can create a ticket to do that in the next iteration of the metrics.

libs/metrics/social_distancing.py
pgrill: If we put the processing time inside the if we will lose the `processing_time` of the first status.

api/routers/metrics/area_metrics.py
```diff
@@ -0,0 +1,269 @@
...
+from api.models.metrics import (
+    FaceMaskDaily, FaceMaskHourly, FaceMaskWeekly, FaceMaskLive, SocialDistancingDaily, SocialDistancingHourly,
+    SocialDistancingWeekly, SocialDistancingLive, OccupancyHourly, OccupancyDaily, OccupancyLive, OccupancyWeekly)
+from api.utils import extract_config
+from libs.metrics import FaceMaskUsageMetric, OccupancyMetric, SocialDistancingMetric
+
+metrics_router = APIRouter()
+
+
+def get_all_areas() -> str:
```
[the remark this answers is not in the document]
pgrill: My idea was to return a string to use as "default" parameters.
However, you are right, I am splitting it always later. I change the function to return a list.

api/routers/metrics/camera_metrics.py
```diff
@@ -0,0 +1,195 @@
...
+    SocialDistancingDaily, SocialDistancingHourly, SocialDistancingLive,
+    SocialDistancingWeekly)
+from api.utils import extract_config
+from libs.metrics import FaceMaskUsageMetric, SocialDistancingMetric
+from libs.metrics.utils import generate_heatmap
+
+metrics_router = APIRouter()
+
+
+def get_all_cameras() -> str:
```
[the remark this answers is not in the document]
pgrill: Changed.

libs/metrics/social_distancing.py
mats-claassen: I mean like this you are overriding the `processing_time` so you will lose the time that elapses from the status change until `processing_count >= cls.processing_count_threshold`, right?

pgrill: :thinking: You are right, but inside the if I will lose the first status.
I need to use a new variable to store the time when the status changes.

pgrill: I just committed a fix.

libs/metrics/face_mask_usage.py
```diff
@@ -0,0 +1,115 @@
...
+        return total_no_face_detections, total_mask_detections, total_no_mask_detections
+
+    @classmethod
+    def generate_live_csv_data(cls, today_entity_csv, entity):
+        """
+        Generates the live report using the `today_entity_csv` file received.
+        """
+        with open(today_entity_csv, "r") as log:
+            objects_logs = {}
+            lastest_entries = deque(csv.DictReader(log), 1000)
```
renzodgc, line 98: Can we compute this 1000 given the frequency with which we update live data and the raw data we log each second?

libs/metrics/occupancy.py
```diff
@@ -0,0 +1,98 @@
...
+        return round(mean(average_ocupancy), 2), max(max_occupancy)
+
+    @classmethod
+    def generate_live_csv_data(cls, today_entity_csv, entity):
+        """
+        Generates the live report using the `today_entity_csv` file received.
+        """
+        with open(today_entity_csv, "r") as log:
+            objects_logs = {}
+            lastest_entries = deque(csv.DictReader(log), 1000)
```
renzodgc, line 60: same comment here

libs/metrics/occupancy.py
```diff
@@ -0,0 +1,98 @@
...
+        reports_directory = os.path.join(entity_directory, "reports", cls.reports_folder)
+        file_path = os.path.join(reports_directory, "live.csv")
+        if os.path.exists(file_path):
+            with open(file_path, "r") as live_file:
+                lastest_entry = deque(csv.DictReader(live_file), 1)[0]
+                if datetime.strptime(lastest_entry["Time"], "%Y-%m-%d %H:%M:%S").date() == datetime.today().date():
+                    daily_violations = int(lastest_entry["Violations"])
+        if occupancy_live[1] > occupancy_live[2]:
+            # Max Occupancy detections > Occupancy threshold
+            daily_violations += 1
```
renzodgc, line 82: This is not daily but in the last few minutes, right?

libs/metrics/social_distancing.py
```diff
@@ -0,0 +1,189 @@
...
+            cls.create_heatmap_report(config, yesterday_csv, violation_heatmap_file, "Violations")
+
+    @classmethod
+    def generate_live_csv_data(cls, today_entity_csv, entity):
+        """
+        Generates the live report using the `today_entity_csv` file received.
+        """
+        with open(today_entity_csv, "r") as log:
+            objects_logs = {}
+            lastest_entries = deque(csv.DictReader(log), 1000)
```
renzodgc, line 172: same here

renzodgc (commented):
Amazing work :raised_hands:

api/routers/export_data.py
```diff
@@ -0,0 +1,95 @@
...
+            export_folder_into_zip(social_ditancing_reports_path, os.path.join(
+                "cameras", f"{cam_id}-{name}", social_ditancing_reports_folder), export_zip)
+            export_folder_into_zip(face_mask_reports_path, os.path.join(
+                "cameras", f"{cam_id}-{name}", face_mask_reports_folder), export_zip)
+        for (area_id, name) in areas:
+            occupancy_logs_path = os.path.join(os.getenv("AreaLogDirectory"), area_id, "occupancy_log")
+            occupancy_report_folder = f"reports/{OccupancyMetric.reports_folder}"
+            occupancy_report_path = os.path.join(os.getenv("AreaLogDirectory"), area_id,
+                                                 occupancy_report_folder)
+            export_folder_into_zip(occupancy_logs_path, os.path.join(
```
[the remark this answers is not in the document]
pgrill: The occupancy log contains the area's raw data.
The reports' ones contain the metrics results.

libs/metrics/base.py
```diff
@@ -0,0 +1,321 @@
...
+            else:
+                # cls.entity == "area"
+                log_directory = os.path.join(entity_directory, "occupancy_log")
+            reports_directory = os.path.join(entity_directory, "reports", cls.reports_folder)
+            # Create missing directories
+            os.makedirs(log_directory, exist_ok=True)
+            os.makedirs(reports_directory, exist_ok=True)
+            if current_hour == 0:
+                # Pending to process the latest hour from yesterday
+                time_until = datetime.combine(date.today(), time(0, 0))
```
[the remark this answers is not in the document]
pgrill: we can write the logic as

```
time_until = datetime.combine(date.today(), time(0, 0))
report_date = date.today() - timedelta(days=1)
if current_hour != 0:
  time_until = datetime.combine(date.today(), time(current_hour, 0))
  report_date = date.today()
```

I think that the `if` ... `else` syntax is more clear.

libs/metrics/base.py
```diff
@@ -0,0 +1,321 @@
...
+            else:
+                # cls.entity == "area"
+                log_directory = os.path.join(entity_directory, "occupancy_log")
+            today_entity_csv = os.path.join(log_directory, str(date.today()) + ".csv")
+            live_report_csv = os.path.join(reports_directory, "live.csv")
+            csv_headers = cls.live_csv_headers if cls.live_csv_headers else cls.csv_headers
+            headers = ["Time"] + csv_headers
+            report_file_exists = os.path.isfile(live_report_csv)
+            if not os.path.isfile(today_entity_csv):
+                return
```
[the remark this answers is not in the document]
pgrill: The live metrics are not updated. So, if you request the `live-feed` endpoint you will receive that previous day data (or zeros if you just started the processor)/

libs/metrics/occupancy.py
pgrill: In that case, the occupancy live endpoints return also the accumulative violations in the day.

libs/metrics/base.py
```diff
@@ -0,0 +1,321 @@
...
+                with open(daily_csv, "r", newline='') as csvfile:
+                    processed_hours = sum(1 for line in csv.reader(csvfile)) - 1
+                    time_from = datetime.combine(report_date, time(processed_hours + 1, 0))
+            else:
+                with open(daily_csv, "a", newline='') as csvfile:
+                    writer = csv.DictWriter(csvfile, fieldnames=cls.csv_headers)
+                    writer.writeheader()
+            csv_data = cls.generate_hourly_csv_data(entity["id"], entity_csv, time_from, time_until)
+            if csv_data is None:
+                continue
```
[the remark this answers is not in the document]
pgrill: Done

libs/metrics/occupancy.py
pgrill: Done

libs/metrics/social_distancing.py
pgrill: Done

libs/metrics/face_mask_usage.py
pgrill: Done

libs/metrics/base.py
```diff
@@ -0,0 +1,321 @@
...
+            # Create missing directories
+            os.makedirs(log_directory, exist_ok=True)
+            os.makedirs(reports_directory, exist_ok=True)
+            if current_hour == 0:
+                # Pending to process the latest hour from yesterday
+                time_until = datetime.combine(date.today(), time(0, 0))
+                report_date = date.today() - timedelta(days=1)
+            else:
+                time_until = datetime.combine(date.today(), time(current_hour, 0))
+                report_date = date.today()
```
renzodgc, line 105: Why not like this? :thinking: 
```suggestion
            time_until = datetime.combine(date.today(), time(current_hour, 0))
            if current_hour == 0:
                # Pending to process the latest hour from yesterday
                report_date = date.today() - timedelta(days=1)
            else:
                report_date = date.today()
```

libs/metrics/occupancy.py
renzodgc: Ah true!

renzodgc (approved):

PR #101, opened 2020-12-16
Openpifpaf and OFM face-mask classifier TensorRT support for Jetson TX2 with Jetpack 4.4

undefined-references:
Changes:
* Openpifpaf_tensorrt added as a detector for Jetson and is tested on Jetson TX2.
* OFM tensorrt added as a face-mask classifier for Jetson and is tested on Jetson TX2.
* Dockerfile modified to nvcr base image 'l4t-pytorch:r32.4.4-pth1.6-py3' which is compatible with jetson's with Jetpack 4.4
* Previous dockerfile (compatible with Jetpack 4.3) **is removed** (with an assumption to have the latest versions compatibility on master, you can create a tag current state for Jetpack 4.3 and then merge this PR).
* Due to some conflicts between openpifpaf 0.12a4 and pytorch 1.6, exporting ONNX model from openpifpaf on Jetson device was not possible, so I have sent this [PR](https://github.com/neuralet/neuralet-models/pull/15) to Neuralet-Models containing openpifpaf version 0.12a4 ONNX models and I need to modify model download url in [generate_tensorrt script](https://github.com/undefined-references/smart-social-distancing/blob/8e440df64682991df118b391da0ed384d66d922c/generate_tensorrt.bash#L18) after merging that PR (before merging this PR).


generate_tensorrt.bash
```diff
@@ -1,25 +1,41 @@
 #!/bin/bash
 config="$1"
+run_on_jetson="$2"
```
JsonSadler, line 3: Make sure this is also changed in the docs and readme

JsonSadler (approved):
Thanks for your great work :)

libs/classifiers/classifier.py
```diff
@@ -1,17 +1,24 @@
+import logging
+
+logging.getLogger().setLevel(logging.INFO)
+
 class Classifier:
```
mrn-mln, line 6: Could you add the docstring for the class?

mrn-mln (approved):

generate_tensorrt.bash
```diff
@@ -1,25 +1,41 @@
...
-if [ ! -f $tensorrtName ]; then
-	echo ############## Generating TensorRT Engine ######################
-	onnx2trt $onnxName -o $tensorrtName -d $precision -b 1
-fi	
+if [ ! -f $tensorrt_name_openpifpaf ]; then
+    echo "############## Generating TensorRT Engine for openpifpaf ######################"
+    onnx2trt $onnx_name_openpifpaf -o $tensorrt_name_openpifpaf -d $precision -b 1
+fi
+
+tensorrt_name_face_mask="/repo/data/tensorrt/ofm_face_mask_d${precision}.trt"
```
pgrill, line 35: Is correct to use the same precision in the detector and the classifier?
Maybe we can add the `TensorrtPrecision` attribute in the classifier section.

libs/detectors/jetson/openpifpaf_tensorrt/openpifpaf_tensorrt.py
```diff
@@ -0,0 +1,212 @@
+import time
```
pgrill, line 1: Is this file a copy of the x86 version?

libs/classifiers/classifier.py
undefined-references: Sure, thanks.

config-jetson.ini
```diff
@@ -64,13 +64,21 @@ DailyReportTime = 06:00
...
 Device = Jetson
 ;  Name can be ssd_mobilenet_v2_coco which is trained on coco, or
-; ssd_mobilenet_v2_pedestrian_softbio trained by neuralet on softbio dataset
-Name = ssd_mobilenet_v2_coco
-;ImageSize should be 3 numbers seperated by commas, no spaces: 300,300,3
-ImageSize = 300,300,3
+; ssd_mobilenet_v2_pedestrian_softbio trained by neuralet on softbio dataset or it can be openpifpaf_tensorrt which is a pose estimation model
+Name = openpifpaf_tensorrt
+;ImageSize should be 3 numbers seperated by commas, no spaces
+ImageSize = 321,193,3
```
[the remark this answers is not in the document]
undefined-references: It's a good idea, thanks.

generate_tensorrt.bash
undefined-references: You are right. I will modify this.

libs/detectors/jetson/openpifpaf_tensorrt/openpifpaf_tensorrt.py
undefined-references: x86? no but yes for x86-gpu-tensorrt
This happens a lot on other places too due to the policy to separate modules based on platforms to have a straightforward structure. There is a solution to change the structure and add one gpu-tensorrt device for both x86-gpu-tensorrt and Jetson devices.

generate_tensorrt.bash
undefined-references: Sure, Thanks

pgrill (approved):
Great work!!!

undefined-references:
@pgrill Thanks :)

PR #102, opened 2020-12-18
Fixed miner issue at classifier module

mrn-mln:
Some detectors doesn't detect face as a result there is no 'face' key at object dictionary which raise error. 
The key is checked before using.
`
ERROR:root:'face'
  File "/repo/libs/classifiers/classifier.py", line 21, in object_post_process
    if object['face'] is not None and classifier_score > self.min_threshold:
KeyError: 'face'
`


mrn-mln:
@pgrill 
@renzodgc

libs/classifiers/classifier.py
```diff
@@ -18,7 +18,8 @@ def inference(self, objects):
         return self.classifier.inference(objects)
 
     def object_post_process(self, object, classifier_result, classifier_score):
-        if object['face'] is not None and classifier_score > self.min_threshold:
-            object['face_label'] = classifier_result
-        else:
-            object['face_label'] = -1
+        if object.get('face') is not None:
```
JsonSadler, line 21: Again, if you intend to check existence, you have to check with `in` operator, not `get`. Explained in your other PR for yolov3. 

Thanks for your bughunt, great work.

JsonSadler (changes requested):

libs/classifiers/classifier.py
mrn-mln: `get` and `in` both work. I'll change it to in

pgrill (approved):
Great work

pgrill (approved):

PR #103, opened 2020-12-19
Yolov3

mrn-mln:
Adding yolov3 detector.


libs/classifiers/classifier.py
```diff
@@ -18,7 +18,8 @@ def inference(self, objects):
         return self.classifier.inference(objects)
 
     def object_post_process(self, object, classifier_result, classifier_score):
-        if object['face'] is not None and classifier_score > self.min_threshold:
-            object['face_label'] = classifier_result
-        else:
-            object['face_label'] = -1
+        if object.get('face') is not None:
```
JsonSadler, line 21: If you check for existence, please do it with `if 'face' in object` rather than using `.get()`, I THINK if someday having None as the value of the key 'face' will be meaningful, this breaks the code as it exists but returns None.

libs/detectors/x86/yolov3.py
```diff
@@ -0,0 +1,108 @@
...
+            print('using %s as model' % user_model_path)
+            self.model_path = user_model_path
+        else:
+            url = 'https://github.com/neuralet/neuralet-models/blob/master/amd64/coco_yolo_v3/yolov3.weights?raw=true'
+
+            if not os.path.isfile(self.model_path):
+                print('model does not exist under: ', self.model_path, 'downloading from ', url)
+                wget.download(url, self.model_path)
+
+        self.nms_thesh = 0.5
```
JsonSadler, line 40: I think you meant `nms_thresh`, missed the `r`.

libs/detectors/x86/yolov3.py
```diff
@@ -0,0 +1,108 @@
...
+
+        result = []
+        for i, pred in enumerate(output):
+            c1 = pred[1:3].cpu().int().numpy()  # unormalized [xmin, ymin]
+            c2 = pred[3:5].cpu().int().numpy()  # unormalized [xmax, ymax]
+            cls = int(pred[-1].cpu())
+            score = float(pred[5].cpu)
+            if cls == 0:  # person class index is '0' at coco dataset
+                bbox_dict = {"id": "1-" + str(i),
+                             "bbox": [c1[1] / self.h, c1[0] / self.w, c2[1] / self.h, c2[0] / self.w], "score": score,
```
JsonSadler, line 105: Is this float division? self.w and self.h are integers.

JsonSadler (changes requested):

libs/detectors/x86/yolov3.py
mrn-mln: c1 and c2 both are `int`. The division works for normalizing the bbox.

config-x86.ini
```diff
@@ -64,11 +64,12 @@ DailyReportTime = 06:00
...
 ; Supported devices: Jetson , EdgeTPU, Dummy, x86
 Device = x86
-; Supported models: mobilenet_ssd_v2 and openpifpaf
-Name = openpifpaf
-;ImageSize should be 3 numbers seperated by commas, no spaces: 300,300,3 (for better accuracy use higher resolution when
+; Supported models: mobilenet_ssd_v2, openpifpaf, and yolov3
+Name = yolov3
+; ImageSize should be 3 numbers seperated by commas, no spaces: 300,300,3 (for better accuracy use higher resolution when
 ; using openpifpaf (openpifpaf detects both faces and pedestrians)
-ImageSize = 1281,721,3
```
mats-claassen, line 71: Could we keep this resolution for OpenPifPaf in a comment?

mhejrati (approved):
LGTM,

PR #106, opened 2020-12-21
Improve s3 logger

pgrill:
Create S3 backup periodic task.


README.md
```diff
@@ -126,7 +130,7 @@ As you are using a self-signed certificate you will need to import the created C
 docker build -f jetson-nano.Dockerfile -t "neuralet/smart-social-distancing:latest-jetson-nano" .
 
 # 3) Run Docker container:
-docker run -it --runtime nvidia --privileged -p HOST_PORT:8000 -v "$PWD":/repo -e TZ=`./timezone.sh` neuralet/smart-social-distancing:latest-jetson-nano
+docker run -it --runtime nvidia --privileged -p HOST_PORT:8000 -v "$PWD":/repo -e TZ=`./timezone.sh` --env-file .env neuralet/smart-social-distancing:latest-jetson-nano
```
mats-claassen, line 133: Adding this flag is optional right? What if the file does not exist, does it work? If not then I would prefer to remove the flag from these commands and just leave a note that when you want to use that feature you must add the flag to the command (as you did in the S3 section)

README.md
```diff
@@ -348,9 +352,12 @@ All the configurations are grouped in *sections* and some of them can vary depen
...
+  The processor also supports the execution of periodic tasks to generate reports, accumulate metrics, backup your files, etc. For now, we support the *metrics* and *s3_backup* tasks. You can enable/disable these functionalities uncommenting/commenting the section or with the *Enabled* flag.
   - `metrics`: Generates different reports (hourly, daily and live) with information about the social distancing infractions, facemask usage and occupancy in your cameras and areas. You need to have it enabled to see data in the [UI](https://beta.lanthorn.ai) dashboard or use the `/metrics` endpoints.
       - `LiveInterval`: Expressed in minutes. Defines the time interval desired to generate live information.
+  - `s3_backup`: Back up into an S3 bucket all the generated data (raw data and reports). To enable the functionality you need to configure the aws credentials following the steps explained in the section *Configuring AWS credentials*
```
mats-claassen, line 358: You could add a link to the section instead of only naming it

README.md
pgrill: You are right. I removed all the optional flag from the suggested commands.

README.md
pgrill: Done

mats-claassen (approved):

PR #112, opened 2020-12-24
Include authentication in endpoints

pgrill:


README.md
```diff
@@ -119,6 +119,40 @@ If you don't have a certificate for the processor, you can create a self-signed
...
+To configure OAuth2 in the processor you need to follow these steps:
+  1. Enabling OAuth2 in the API by setting in `True` the parameter `UseAuthToken` (included in the `API` section).
+  2. Set into the container the env `SECRET_ACCESS_KEY`. This env is used to encode the JWT token. An easy way to do that is 
+     to create a `.env` file (following the template `.env.example`) and pass the flag ```--env-file .env ``` when you run the processor.
+  3. Create an API user. You can do that in two ways:
+     1. Using the `create_api_user.py` script:
+
+      Inside the docker container, execute the script `python3 create_api_user.py --user=<USER> --password=<PASSWORD>`. For example, if you are using an x86 device, you can execute the following script.
+      ```bash
+      docker run -it -p HOST_PORT:8000 -v "$PWD":/repo -e TZ=`./timezone.sh` neuralet/smart-social-distancing:latest-x86_64 python3 create_api_user.py --user=<USER> --password=<PASSWORD>
```
mats-claassen, line 135: Would you have to run this every time you restart the processor?

README.md
```diff
@@ -119,6 +119,40 @@ If you don't have a certificate for the processor, you can create a self-signed
...
+     2. Using the `/auth/create_api_user` endpoint:
+      Send a POST request to the endpoint `http://<PROCESSOR_HOST>:<PROCESSOR_PORT>/auth/create_api_user` with the following body:
+      ```
+      {
+          "user": <USER>,
+          "password": <PASSWORD>
+      }
+      ```
+
+      After executing one of these steps, the `user` and `password` (hashed) will be stored in the file `/repo/data/auth/api_user.txt` inside the container. To avoid lost that file when the container is restarted, we recommend mounting the `/repo` directory.
```
mats-claassen, line 146:
```suggestion
      After executing one of these steps, the `user` and `password` (hashed) will be stored in the file `/repo/data/auth/api_user.txt` inside the container. To avoid losing that file when the container is restarted, we recommend mounting the `/repo` directory as a volume.
```

Also, mounting `data` to `/repo/data` would be enough right?

api/processor_api.py
```diff
@@ -97,8 +104,6 @@ async def validation_exception_handler(request: Request, exc: RequestValidationE
             app.add_middleware(CORSMiddleware, allow_origins="*", allow_credentials=True, allow_methods=["*"],
                                allow_headers=["*"])
 
-        app.mount("/static", StaticFiles(directory="/repo/data/processor/static"), name="static")
```
mats-claassen, line 100: Without this, can the frontend still access the files in this directory? I don't think so. For example the video files. The csv files are not used directly anymore?

libs/utils/auth.py
```diff
@@ -0,0 +1,54 @@
...
+def create_access_token(user: str):
+    # Set a week (60*24*7=10080) as expiration date
+    expire_date = datetime.utcnow() + timedelta(minutes=10080)
+    data = {"sub": user}
+    data.update({"exp": expire_date})
+    return jwt.encode(data, os.environ.get("SECRET_ACCESS_KEY"), algorithm="HS256")
+
+
+def validate_jwt_token(token):
+    jwt.decode(token, os.environ.get("SECRET_ACCESS_KEY"), algorithms=["HS256"])
```
mats-claassen, line 54: Is calling decode enough or should we validate some of the content as well?

README.md
pgrill: No, only the first time.

api/processor_api.py
pgrill: I changed the `StaticFiles` into a router to include the authentication middleware (api/routers/static.py).
After some research, I did;t find any way to force authentication using `StaticFiles `.

libs/utils/auth.py
pgrill: We can also validate that the token sent is attached to the user created.
I will include that validation.

api/routers/auth.py
```diff
@@ -0,0 +1,30 @@
...
+
+@auth_router.post("/create_api_user", status_code=status.HTTP_204_NO_CONTENT)
+async def create_user(auth_info: AuthDTO):
+    """
+    Creates the API user (if it's not already created) with the given password.
+    """
+    try:
+        create_api_user(auth_info.user, auth_info.password)
+    except Exception as e:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
```
renzodgc, line 18: Shouldn't we raise this inside create_api_user? or is not ideal because another script calls this outside of the API?

pgrill: The `create_api_user` script can be called directly for the user. In that case doesn't make sense return a HTTP status

renzodgc (approved):

libs/utils/auth.py
pgrill: I included the user validation in the commit 9872ce64320acea947fbaf71dd55f6d5ed7209bc

mats-claassen (approved):

PR #114, opened 2020-12-29
Improve export endpoints

pgrill:
The export data endpoint was improved adding some filtering features.
The current endpoint allows the user to select the data to export taking filtering by:
- *Entity*: (areas or cameras).
- *Dates*: (only include data for the specified date range).
- *Data Type*: (the type of information that you want to export. The available values are raw_data, occupancy, social-distancing, facemask-usage and all_data)


api/routers/export.py
```diff
@@ -0,0 +1,169 @@
...
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail=f"Areas with ids {missing_areas} don't exist."
+            )
+    if selected_areas:
+        return [(area["Id"], area["Name"]) for area in selected_areas]
+    return []
+
+
+def get_cameras(export_info: ExportDTO, areas: List[Tuple[str, str]]) -> List[Tuple[str, str]]:
```
mats-claassen, line 64: This function exports data from cameras in export_info and cameras in `areas` but it is not immediately obvious what it does. Maybe a short doctring could explain it. I see that we aren't adding them to other endpoints but it would still be good at least when the function is not completely obvious

pgrill: I include the docstring in the commit 31ec50682a5b52bc597a408b64508660880fbcc9

api/routers/export.py
```diff
@@ -0,0 +1,169 @@
...
+        selected_cameras.extend(
+            [c for c in all_cameras if c["Id"] in areas_cameras and c["Id"] not in export_info.cameras]
+        )
+    if selected_cameras:
+        return [(camera["Id"], camera["Name"]) for camera in selected_cameras]
+    return []
+
+
+def export_camera_data_into_file(export_info: ExportDTO, camera_id: str, camera_name: str, zip_file: str) -> None:
+    if ALL_DATA in export_info.data_types or RAW_DATA in export_info.data_types:
```
renzodgc, line 107: I may be mistaken, but shouldn't inclusion be validated in the opposite way?

api/routers/export.py
```diff
@@ -0,0 +1,169 @@
...
+                date_matches = re.findall(r"([0-9]{4}\-[0-9]{2}\-[0-9]{2})", filename)
+                if not date_matches:
+                    continue
+                file_date = datetime.strptime(date_matches[0], '%Y-%m-%d').date()
+                if not from_date <= file_date <= to_date:
+                    continue
+            zip_file.write(os.path.join(source_path, filename), arcname=os.path.join(destination_path, filename))
+
+
+def get_areas(export_info: ExportDTO) -> List[Tuple[str, str]]:
```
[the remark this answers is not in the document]
pgrill: :thinking: these methods don't retrieve the area's (or camera) data. They return a list with the (id, name) of the requested areas and cameras.

api/routers/export.py
pgrill: :thinking: What do you mean by "the opposite way"?

api/routers/export.py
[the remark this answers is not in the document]
mats-claassen: These functions are only used in this file, right? In that case I on't think it is toobad to repeat function names. I think when a function is internal to a file we could prefix them with _

mats-claassen (approved):

api/routers/export.py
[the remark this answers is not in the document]
renzodgc: Yeah my bad, I think after that many code reviews yesterday I got mixed up :sweat_smile:
Either an underscore or a name as `get_areas_from_export_info` would be good for me

api/routers/export.py
renzodgc: Nevermind, I just reread the code and understood the constants were strings :upside_down_face: I don't know why yesterday I thought they were lists or something, maybe the same reason as my previous comment :sweat_smile:

api/routers/export.py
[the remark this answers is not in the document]
pgrill: I renamed the functions as `get_areas_to_export` and  `get_cameras_to_export`.
I agree with the usage of the `_`. However, it's not used in the rest of the repository and I think that will be strange use it in a single file.

The name `get_areas_from_export_info` doen't work for the cameras because it also returns the cameras included in the selected areas.

renzodgc (approved):

PR #117, opened 2020-12-31
Fix master for jetson nano

pgrill:
The master branch is broken for jetson-nano because the default jetson-config file use Openpifpaf models build with JetPack 4.4.

The current jetson-nano dockerfile doesn't work because it doesn't include `torch`. Moreover, the `sd_mobilenet_v2_coco` and `pedestrian_ssd_mobilenet_v2_softbio` models downloaded by the script `download_jetson_nano_trt.sh` are built using JetPack 4.3.

I split the `config-jetson.ini` into 2 different files, one for each device.


mats-claassen:
What is missing to have both Nano and TX2 working on Jetpack 4.4?

pgrill:
> What is missing to have both Nano and TX2 working on Jetpack 4.4?

We need to change the `mobilenet_v2_coco` and `mobilenet_v2_pedestrian_softbio` models for jetson nano. If we want also to support Openpifpaf in jetson nano we need to change the dockerfile (the current one doesn't have included torch).

I just create the PR because if we create a release with the master the jetson nano device will not work.
Maybe we can open a new issue to include Jetpack 4.4 for the jetson-nano device.

mats-claassen:
Creating an issue would be good.

pgrill:
> Creating an issue would be good.

I created the issue https://github.com/neuralet/smart-social-distancing/issues/119

renzodgc (approved):

mats-claassen (approved):

PR #120, opened 2021-01-04
Api improvements

pgrill:
Small fixes in the processor API

- The response of the endpoints in the PUT and POST operation was changed to keep the GET format.
- The metrics endpoints were refactored to avoid repeating code.

### DIFF
--- a/api/routers/metrics/area_metrics.py
+++ b/api/routers/metrics/area_metrics.py
@@ -1,50 +1,12 @@
-import os
-
 from datetime import date, timedelta
-from fastapi import APIRouter, Query, HTTPException, status
-from typing import Iterator
+from fastapi import APIRouter, Query
 
 from api.models.metrics import (
     FaceMaskDaily, FaceMaskHourly, FaceMaskWeekly, FaceMaskLive, SocialDistancingDaily, SocialDistancingHourly,
     SocialDistancingWeekly, SocialDistancingLive, OccupancyHourly, OccupancyDaily, OccupancyLive, OccupancyWeekly)
-from api.utils import bad_request_serializer, extract_config
-from libs.metrics import FaceMaskUsageMetric, OccupancyMetric, SocialDistancingMetric
+from .metrics import get_live_metric, get_hourly_metric, get_daily_metric, get_weekly_metric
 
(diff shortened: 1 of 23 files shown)


api/routers/metrics/area_metrics.py
```diff
@@ -58,10 +24,7 @@ def get_area_occupancy_hourly_report(areas: str = "", date: date = Query(date.to
     Returns a hourly report (for the date specified) with information about the occupancy in
     the areas <areas>.
     """
-    areas = get_areas(areas)
-    for area in areas:
-        validate_area_existence(area)
-    return OccupancyMetric.get_hourly_report(areas, date)
+    return get_hourly_metric("areas", areas, "occupancy", date)
```
mats-claassen, line 27: For all these constant strings I would prefer if we have them as constants somewhere and use those constants instead or repeating the string everywhere. I think we have a constants file, right? But it could go elsewhere

pgrill: Agree. I didn't include them in the constant file because it's not used in the rest of the project (device names, model names, etc are always strings). However, put them in the file make more sense.

I will change it.

renzodgc (approved):
Great work :clap:

api/routers/metrics/area_metrics.py
```diff
@@ -4,18 +4,20 @@
...
     SocialDistancingWeekly, SocialDistancingLive, OccupancyHourly, OccupancyDaily, OccupancyLive, OccupancyWeekly)
+from constants import AREAS, FACEMASK_USAGE, OCCUPANCY, SOCIAL_DISTANCING
+
 from .metrics import get_live_metric, get_hourly_metric, get_daily_metric, get_weekly_metric
 
 metrics_router = APIRouter()
 
 
-# Occupancy Metrics
+# Occupancy MetricsAREAS
```
renzodgc, line 14:
```suggestion
# Occupancy Metrics
```

api/routers/metrics/area_metrics.py
pgrill: Done

mats-claassen (approved):

PR #121, opened 2021-01-06
Standardize endpoint error messages

renzodgc:
- Changed all errors to constants
- Changed Bad Requests errors to use the same format as fast-api model error serializer


api/utils.py
```diff
@@ -63,21 +63,24 @@ def update_config(config_dict, reboot_processor):
...
+        "loc": loc,
+        "msg": msg,
+        "type": error_type
+    }]
 
 def handle_response(response, success, status_code=status.HTTP_200_OK):
     if not success:
-        return JSONResponse(
+        raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
```
pgrill, line 76: Taking into account that we are not sure about the reason of the error i think that a status 500 is better.
What do you think?

api/utils.py
```diff
@@ -63,21 +63,24 @@ def update_config(config_dict, reboot_processor):
...
-                "type": "unknown error on the config file",
-                "body": humps.decamelize(response)
-            })
+            detail=bad_request_serializer("Failed to restart video processor", error_type="config unknown error")
         )
     content = humps.decamelize(response) if response else None
     return JSONResponse(status_code=status_code, content=content)
 
 
+
```
pgrill, line 83: I think that this empty  line was added by mistake.

mats-claassen:
To make creating Release Notes easier each PR title should be ideally the same entry we would put in the release notes. I think "Refactor endpoint error messages" makes sense but maybe it could be a bit more descriptive. Maybe "Standardize"?

mats-claassen (approved):

api/utils.py
renzodgc: Hmm but it wasn't exactly an error due to server issues right?

pgrill: A bad request error means that the user sent something wrong in the request; for example, he forgot to send a mandatory field. All these errors should be captured by the pydantic models (before call this function).

If we arrive into this function with an error means that for any reason (that we don't know) the server couldn't be restarted. Is common to return a 500 when you have an unexpected error in your system.

api/utils.py
```diff
@@ -63,21 +63,24 @@ def update_config(config_dict, reboot_processor):
...
-            })
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Failed to restart video processor, unknown error on config file"
         )
     content = humps.decamelize(response) if response else None
     return JSONResponse(status_code=status_code, content=content)
 
 
+
```
renzodgc, line 83:
```suggestion

```

pgrill (approved):

PR #123, opened 2021-01-12
Measure performance in logs

pgrill:


README.md
```diff
@@ -309,6 +309,14 @@ All the configurations are grouped in *sections* and some of them can vary depen
...
   - `MaxThreadRestarts`: Defines the number of restarts allowed per thread.
   - `HeatmapResolution`: Sets the resolution used by the heatmap report.
+  - `LogPerformance`: A boolean parameter to enable/disable the logging of "Performance Metrics" in the default processor log.
+  We recommend enabling it to compare the performance of different devices, models, resolutions, etc.
+  When it's enabled, the processor logs will include the following information every time 100 frames are processed:
+      - Frames per second (FPS)
+      - Average Detector time
+      - Average Classifier time
+      - Average Tracker time
+      - Average Post Processors time
```
renzodgc, line 319:
```suggestion
      - Average Post Processing time
```

libs/cv_engine.py
```diff
@@ -2,13 +2,20 @@
...
 from libs.classifiers.classifier import Classifier
 from libs.trackers.tracker import Tracker
 from libs.loggers.source_loggers.logger import Logger
 from libs.detectors.detector import Detector
 from libs.source_post_processors.source_post_processor import SourcePostProcessor
 
+
 logger = logging.getLogger(__name__)
+FRAMES_LOG_BATCH_SIZE = 100
+LOG_SECTIONS = ["Detector", "Classifier", "Tracker", "Post Processors"]
```
renzodgc, line 18: I'd suggest we change each instance of `Post Processors` to `Post Processing`, I think it makes more sense grammatically

pgrill: I changed the logging information for the post-processing steps.
Now, each step is log independently.

README.md
```diff
@@ -309,6 +309,17 @@ All the configurations are grouped in *sections* and some of them can vary depen
   - `OccupancyAlertsMinInterval`:  Sets the desired interval (in seconds) between occupancy alerts.
   - `MaxThreadRestarts`: Defines the number of restarts allowed per thread.
   - `HeatmapResolution`: Sets the resolution used by the heatmap report.
+  - `LogPerformance`: A boolean parameter to enable/disable the logging of "Performance Metrics" in the default processor log.
```
mats-claassen, line 312: What about calling this `LogPerformanceMetrics`?

libs/cv_engine.py
```diff
@@ -50,29 +62,54 @@ def __process(self, cv_image):
...
                     idx = idx + 1
                 else:
                     self.classifier.object_post_process(obj, None, None)
+                classifier_time += (datetime.now() - begin_time).total_seconds()
 
         # Execute post processors
         post_processing_data = {
             "tracks": tracks
         }
+        begin_time = datetime.now()
```
mats-claassen, line 98: You can remove this line

libs/cv_engine.py
```diff
@@ -108,3 +146,28 @@ def process_video(self, video_uri):
 
     def stop_process_video(self):
         self.running_video = False
+
+    def reset_log_detail(self):
+        self.log_detail = {}
+        for section in LOG_SECTIONS:
+            if section == "Post processing steps":
```
mats-claassen, line 153: Can we put this string in a constant at the top of this file?

mats-claassen (commented):
Just thinking whether it would be good to store these metrics in a file. We could name it <device-name>-<model>.txt. I think you will want to compare different models and/or devices and if they are just printed to the console then you will have to copy them manually somewhere to compare. Thoughts?

pgrill:
> Just thinking whether it would be good to store these metrics in a file. We could name it -.txt. I think you will want to compare different models and/or devices and if they are just printed to the console then you will have to copy them manually somewhere to compare. Thoughts?

Yes, I thought about store it in a file. My only concern was that storing it into a file can affect the processor performance for the file access. But I can implement a short version and test it.

README.md
pgrill: Agree, I changed

libs/cv_engine.py
pgrill: Removed

libs/cv_engine.py
pgrill: Done

pgrill:
> > Just thinking whether it would be good to store these metrics in a file. We could name it -.txt. I think you will want to compare different models and/or devices and if they are just printed to the console then you will have to copy them manually somewhere to compare. Thoughts?
> 
> Yes, I thought about store it in a file. My only concern was that storing it into a file can affect the processor performance for the file access. But I can implement a short version and test it.

I added now the possibility of storing the performance metrics logs into a CSV file.

mats-claassen (approved):

renzodgc (approved):

PR #126, opened 2021-01-15
Documented video formats that the processor is compatible with

renzodgc:


README.md
```diff
@@ -280,6 +280,17 @@ To configure OAuth2 in the processor you need to follow these steps:
       ```
       The obtained token will be valid for 1 week (then a new one must be requested from the API) and needs to be sent as an `Authorization` header in all the requests. If you don't send the token (when the `UseAuthToken` attribute is set in `True`), you will receive a `401 Unauthorized` response.
 
+### Video feeds the processor can consume from
```
pgrill, line 283: What about calling this section "Supported video formats"? or "Supported video feeds formats"?

README.md
```diff
@@ -280,6 +280,17 @@ To configure OAuth2 in the processor you need to follow these steps:
...
+This processor uses [OpenCV VideoCapture](https://docs.opencv.org/3.4/d8/dfe/classcv_1_1VideoCapture.html) , which means that it can process:
+
+* Video files that are compatible with [FFmpeg](https://en.wikipedia.org/wiki/FFmpeg)
+* Any URL of video stream in a **public protocol** such as RTSP and HTTP (`protocol://host:port/script_name?script_params|auth`)
+
+Please note that:
+
+* Although this processor can read and process a video, this is mostly a development functionality; this is due to the fact that loggers yield statistics that are time dependant that assume a real-time stream being processed, in which if the processing capacity is lower than the FPS frames are lost in favour of processing new frames. With a video file all frames are processed and on a slower model this might take a while (and yield wrong analytics).
+* Some IP cameras implement their own private protocol that's not compatible with OpenCV.
```
pgrill, line 292: Maybe we can add a comment that in these cases you should check with the camera provider if the device supports exporting the stream in a "public protocol". For example, wyze doesn't support rtsp as default, but you have the possibility of installing a  firmware that supports it.

pgrill (approved):

README.md
```diff
@@ -280,6 +280,21 @@ To configure OAuth2 in the processor you need to follow these steps:
...
+
+If wanting to integrate an IP camera that uses a private protocol, you should check with the camera provider if the device supports exporting its stream in a public protocol.
```
mats-claassen, line 294:
```suggestion
If you want to integrate an IP camera that uses a private protocol, you should check with the camera provider if the device supports exporting its stream in a public protocol.
```

README.md
```diff
@@ -280,6 +280,21 @@ To configure OAuth2 in the processor you need to follow these steps:
...
 
+### Supported video feeds formats
+This processor uses [OpenCV VideoCapture](https://docs.opencv.org/3.4/d8/dfe/classcv_1_1VideoCapture.html), which means that it can process:
+
+* Video files that are compatible with [FFmpeg](https://en.wikipedia.org/wiki/FFmpeg)
+* Any URL of video stream in a **public protocol** such as RTSP and HTTP (`protocol://host:port/script_name?script_params|auth`)
+
+Please note that:
+
+* Although this processor can read and process a video, this is mostly a development functionality; this is due to the fact that loggers yield statistics that are time dependant that assume a real-time stream being processed, in which if the processing capacity is lower than the FPS frames are lost in favour of processing new frames. With a video file all frames are processed and on a slower model this might take a while (and yield wrong analytics).
```
mats-claassen, line 291:
```suggestion
* Although this processor can read and process a video file, this is mostly a development functionality; this is due to the fact that loggers yield statistics that are time dependant that assume a real-time stream being processed, in which if the processing capacity is lower than the FPS, frames are lost in favour of processing new frames. With a video file all frames are processed and on a slower model this might take a while (and yield wrong analytics).
```

mats-claassen (approved):

PR #127, opened 2021-01-15
Autogenereted Ids for Cameras

JFer11:
In this branch, the way camera IDs are generated has been changed.
Now the ID field is no longer required. If it is passed, the endpoint will save it as usual. However, if not, the endpoint will look up all other numeric ID and assign the corresponding ID to the new camera.

### DIFF
--- a/api/models/base.py
+++ b/api/models/base.py
@@ -14,7 +14,7 @@ class SnakeModel(BaseModel):
 
 
 class EntityConfigDTO(SnakeModel):
-    id: Optional[str] = Field(example='SABORR')
+    id: Optional[str] = Field(example='0')
     name: str = Field(example='Kitchen')
 
 
--- a/api/models/base.py
+++ b/api/models/base.py
@@ -14,7 +14,7 @@ class SnakeModel(BaseModel):
 
 
 class EntityConfigDTO(SnakeModel):
-    id: str = Field(example='0')
+    id: Optional[str] = Field(example='SABORR')
     name: str = Field(example='Kitchen')
 
 
--- a/api/routers/cameras.py
+++ b/api/routers/cameras.py
@@ -143,6 +143,35 @@ async def get_camera(camera_id: str):
     return camera
 
 
+def get_first_unused_id(arr):
+    if not arr:
+        return 0
+
+    ids_numbers = []
+    for value in arr:
+        try:
+            number = int(value)
+            if number >= 0:
+                ids_numbers.append(number)
+        except ValueError:
+            pass
+
+    if not ids_numbers:
+        return 0
+
+    ids_numbers.sort()
+
+    result = 0
+    for i in range(0, ids_numbers[len(ids_numbers)-1]+1):
+        if ids_numbers[i] != i:
+            result = i
+            break
+    else:
+        result = ids_numbers[len(ids_numbers)-1] + 1
+
+    return result
+
+
 @cameras_router.post("", response_model=CameraDTO, status_code=status.HTTP_201_CREATED)
 async def create_camera(new_camera: CameraDTO, reboot_processor: Optional[bool] = True):
     """
@@ -151,7 +180,11 @@ async def create_camera(new_camera: CameraDTO, reboot_processor: Optional[bool] 
     config_dict = extract_config()
     cameras_name = [x for x in config_dict.keys() if x.startswith("Source_")]
     cameras = [map_camera(x, config_dict) for x in cameras_name]
-    if new_camera.id in [camera["id"] for camera in cameras]:
+
+    if new_camera.id is None:
+        ids = [camera["id"] for camera in cameras]
+        new_camera.id = get_first_unused_id(ids)
+    elif new_camera.id in [camera["id"] for camera in cameras]:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=bad_request_serializer("Camera already exists", error_type="config duplicated camera")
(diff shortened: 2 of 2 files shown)


api/models/base.py
```diff
@@ -14,7 +14,7 @@ class Config:
 
 
 class EntityConfigDTO(SnakeModel):
-    id: str = Field(example='0')
+    id: Optional[str] = Field(example='SABORR')
```
pgrill, line 17: Can we keep the "0" as example?

api/routers/cameras.py
```diff
@@ -143,6 +143,35 @@ async def get_camera(camera_id: str):
     return camera
 
 
+def get_first_unused_id(arr):
+    if not arr:
```
pgrill, line 147: I think that the `arr` parameter should be renamed as `cameras_ids` (or another descriptive name)
Also, I think that this logic can be simpler if we use regular expressions.
For example, something like:
```
if not cameras_ids:
    return 0
is_a_number = re.compile("^[0-9]+$")
number_ids = [int(id) for id in cameras_ids if is_a_number.match(id)]
if not number_ids:
    return 0
return max(number_ids) + 1
```

JFer11: I agree with most of the recommendations here, however, i want to highlight one problem. If a user manually sets the id=40, the code will suggest the next id to be 41, and maybe there are many id's available before 40.
Should i keep these part of the code:
```
ids_numbers.sort()
result = 0
for i in range(0, ids_numbers[len(ids_numbers)-1]+1):
    if ids_numbers[i] != i:
        result = i
        break
else:
    result = ids_numbers[len(ids_numbers)-1] + 1

return result
```

api/models/base.py
JFer11: Absolutely!

api/routers/cameras.py
pgrill: Yes, it could work. Maybe you can remove the `break` using instead a return.
Something like:
```
for i in range(0, ids_numbers[len(ids_numbers)-1]+1):
    if ids_numbers[i] != i:
        return i
return ids_numbers[len(ids_numbers)-1] + 1
```

What do you think?

JFer11: Of course, I think it is a good idea. I will change it!

api/routers/cameras.py
```diff
@@ -143,6 +144,23 @@ async def get_camera(camera_id: str):
     return camera
 
 
+def get_first_unused_id(cameras_ids):
```
mats-claassen, line 147: So camera_ids are the existing camera ids? I think that is not obvious from the naming. It would be good to name it `existing_camera_ids` or just add a docstring explaining what the function does and what the parameter contains

api/routers/cameras.py
```diff
@@ -143,6 +144,23 @@ async def get_camera(camera_id: str):
...
+    ids_numbers = [int(value) for value in cameras_ids if is_a_number.match(value)]
+
+    if not ids_numbers:
+        return 0
+
+    ids_numbers.sort()
+    for i in range(0, ids_numbers[len(ids_numbers)-1]+1):
+        if ids_numbers[i] != i:
+            return i
+    return ids_numbers[len(ids_numbers)-1] + 1
```
mats-claassen, line 161: Can't you do `ids_numbers[-1] + 1` or is that just with numpy?

pgrill (approved):

mats-claassen (commented):
I wonder what we should do with previously used IDs. With this implementation if I remove camera 0 and then add another camera it will get the same ID, inheriting all the files from the previous camera 0. I guess the most correct thing would be to remove the camera's folder when it is deleted, which I think we aren't doing now. But we would loose all its info as well. Maybe the ID could be derived from the URL so that if a camera is removed and re-added we could potentially "recover" the old stats. But this supposes that camera URLs don't change

api/routers/cameras.py
renzodgc: I think you can

pgrill: Yes, you can

pgrill:
> I wonder what we should do with previously used IDs. With this implementation if I remove camera 0 and then add another camera it will get the same ID, inheriting all the files from the previous camera 0. I guess the most correct thing would be to remove the camera's folder when it is deleted, which I think we aren't doing now. But we would loose all its info as well. Maybe the ID could be derived from the URL so that if a camera is removed and re-added we could potentially "recover" the old stats. But this supposes that camera URLs don't change

Yes, I agree with you about the risk of "move data" between cameras when you delete/add cameras. However, I am not sure if we should fix it in the context of this PR or we can create a new one.

api/routers/cameras.py
```diff
@@ -158,7 +158,7 @@ def get_first_unused_id(cameras_ids):
     for i in range(0, ids_numbers[len(ids_numbers)-1]+1):
         if ids_numbers[i] != i:
             return i
-    return ids_numbers[len(ids_numbers)-1] + 1
+    return ids_numbers[-1] + 1
```
mats-claassen, line 161: Same applies for line 158 and maybe some others before(?)

mats-claassen:
> > I wonder what we should do with previously used IDs. With this implementation if I remove camera 0 and then add another camera it will get the same ID, inheriting all the files from the previous camera 0. I guess the most correct thing would be to remove the camera's folder when it is deleted, which I think we aren't doing now. But we would loose all its info as well. Maybe the ID could be derived from the URL so that if a camera is removed and re-added we could potentially "recover" the old stats. But this supposes that camera URLs don't change
> 
> Yes, I agree with you about the risk of "move data" between cameras when you delete/add cameras. However, I am not sure if we should fix it in the context of this PR or we can create a new one.

We should add a ticket for it then

renzodgc (approved):

PR #130, opened 2021-01-21
Tests camera routers

JFer11:
In this branch, camera endpoints tests were added.
It was developed by continuing the work of the tests_api_processor_routers branch


api/tests/utils/common_functions.py
```diff
@@ -129,3 +141,23 @@ def camel_case_to_snake_case_dict(dictionary):
...
+        "id": '49',
+        "name": "Kitchen",
+        "video_path": "/repo/data/softbio_vid.mp4",
+        "tags": "kitchen,living_room",
+        "dist_method": "CenterPointsDistance",
+        "live_feed_enabled": True
+    }
+    camera_sample = example
+    return camera_sample
+"""
```
mats-claassen, line 163: Please add a new line at the end of the file

api/tests/utils/fixtures_tests.py
```diff
@@ -33,3 +49,59 @@ def config_rollback():
...
+
+@pytest.fixture
+def camera_sample():
+    camera_sample = copy.deepcopy(example)
+    return camera_sample
+
+
+@pytest.fixture
+def rollback_screenshot_camera_folder():
+    yield None
```
mats-claassen, line 62: What is the purpose of this?

api/tests/utils/fixtures_tests.py
```diff
@@ -33,3 +49,59 @@ def config_rollback():
...
+    if os.path.exists(camera_screenshot_directory):
+        shutil.rmtree(camera_screenshot_directory)
+
+
+@pytest.fixture
+def rollback_homography_matrix_folder():
+    yield None
+    # Deletes the homography_matrix directory and all its content.
+    raw = f'/repo/data/processor/static/data/sources/'
+    path = os.path.join(raw, str(example['id']))
```
mats-claassen, line 74: This will clear all the data for the camera, not just the homography matrix, right?

api/tests/app/test_camera.py
```diff
@@ -0,0 +1,645 @@
...
+import numpy
+
+from api.tests.utils.common_functions import get_section_from_config_file, get_config_file_json
+
+# The line below is absolutely necessary. Fixtures are passed as arguments to test functions. That is why IDE could
+# not recognized them.
+from api.tests.utils.fixtures_tests import config_rollback, camera_sample, rollback_screenshot_camera_folder, h_inverse_matrix, pts_destination, rollback_homography_matrix_folder
+
+"""
+TO DO: Test stuffs related with the parameter reboot_processor.
```
mats-claassen, line 14: The format for a TODO is normally like this:

```python
# TODO: some message
```

This will be highlighted in some IDEs and is easier to search for.

BTW, just to make sure, is this TODO still valid?

api/tests/app/test_camera.py
```diff
@@ -0,0 +1,645 @@
...
+    list_of_cameras = {
+        'cameras': [config_file_json[camera] for camera in config_file_json if camera.startswith('source__')]}
+
+    for camera in list_of_cameras['cameras']:
+        camera.update({'id': str(camera['id'])})
+
+        if with_image:
+            from api.routers.cameras import get_camera_default_image_string
+            image_string = get_camera_default_image_string(camera['id'])
+            # Do not forget to decode it here, to turn into string.
```
mats-claassen, line 47: I don't think this comment is necessary

api/tests/app/test_camera.py
```diff
@@ -0,0 +1,645 @@
...
+def get_camera_from_config_file(camera_id, config_sample_path):
+    list_of_cameras = get_all_cameras(config_sample_path)['cameras']
+    for camera in list_of_cameras:
+        if camera['id'] == camera_id:
+            return camera
+    return None
+
+
+# pytest -v api/tests/app/test_camera.py::TestClassListCameras
+class TestClassListCameras:
```
mats-claassen, line 64: Not sure about conventions here but I think the "Class" part is not necessary. What I know about tests is that they are normally called like `TestListCameras`

mats-claassen: https://docs.pytest.org/en/reorganize-docs/new-docs/user/naming_conventions.html

api/tests/app/test_camera.py
```diff
@@ -0,0 +1,645 @@
...
+        assert response.status_code == 204
+        assert get_h_inverse(camera_id) == h_inverse_matrix['h_inverse.txt']
+
+    # pytest -v api/tests/app/test_camera.py::TestClassConfigCalibratedDistance::test_try_set_coordinates_0_arrays
+    def test_try_set_coordinates_0_arrays(self, config_rollback, camera_sample, rollback_screenshot_camera_folder, rollback_homography_matrix_folder):
+        client, config_sample_path = config_rollback
+        create_a_camera(client, camera_sample)
+
+        body = {
+            "pts_destination": [
```
mats-claassen, line 468: I see a mixture of double and single quotes in this file. I am sure we want to use double quotes for docstrings and I think we agreed to use single quotes for the rest. @renzodgc do you remember?

api/tests/app/test_camera.py
```diff
@@ -0,0 +1,645 @@
...
+"""
+TO DO: Test stuffs related with the parameter reboot_processor.
+"""
+
+
+def create_a_camera(client, camera):
+    return client.post('/cameras', json=camera)
+
+
+def create_2_cameras(client, camera_base, enable=False):
```
renzodgc, line 22: I think we should change this to use the `create_a_camera` method. Also, I dislike limiting it to two.

I'd recommend passing a number of cameras to be created and merging the previous PR

IDs should be handled by the previous merged feature, and the enable logic should be in `create_a_camera` 

The endpoint would return the list of cameras (maybe `create_a_camera`  should return  the params and the response)

api/tests/app/test_camera.py
```diff
@@ -0,0 +1,645 @@
...
+                assert response.json()[key] == camera_sample[key]
+
+    def test_try_create_camera_twice(self, config_rollback, camera_sample, rollback_screenshot_camera_folder):
+        client, config_sample_path = config_rollback
+
+        response_1 = client.post('/cameras', json=camera_sample)
+        response_2 = client.post('/cameras', json=camera_sample)
+
+        assert response_1.status_code == 201
+        assert response_2.status_code == 400
```
renzodgc, line 109: i think these should validate that the nature of the error was that a camera with that id already existed

api/tests/app/test_camera.py
renzodgc: We agreed on double-quotes for everything. So any single quotes should be replaced with double-quotes.

api/tests/app/test_camera.py
```diff
@@ -0,0 +1,645 @@
...
+        assert response.json() == {'detail': [{'loc': ['body', 'pts_destination'], 'msg': 'ensure this value has at '
+                                                                                          'least 4 items',
+                                               'type': 'value_error.list.min_items', 'ctx': {'limit_value': 4}}],
+                                   'body': {'pts_destination': [None]}}
+
+
+# pytest -v api/tests/app/test_camera.py::TestClassGetCameraCalibrationImage
+class TestClassGetCameraCalibrationImage:
+    """ Get Camera Calibration Image, GET /cameras/{camera_id}/calibration_image """
+    """ TODO: Maybe, here we could go deeper into the response.json() """
```
renzodgc, line 535: Do you mean when validating error messages? I think you did it in a test of this class

api/tests/app/test_camera.py
```diff
@@ -0,0 +1,645 @@
...
+        camera_from_config_file_0 = get_camera_from_config_file(camera_id, config_sample_path)
+        camera_from_config_file_1 = get_camera_from_config_file(camera_1['id'], config_sample_path)
+        camera_from_config_file_2 = get_camera_from_config_file(camera_2['id'], config_sample_path)
+
+        assert response.status_code == 204
+        assert camera_from_config_file_0['live_feed_enabled'] is True
+        assert camera_from_config_file_1['live_feed_enabled'] is False
+        assert camera_from_config_file_2['live_feed_enabled'] is False
+
+    def test_enable_one_video_live_feed_disable_false(self, config_rollback, camera_sample,
```
renzodgc, line 629:
```suggestion
    def test_enable_video_feed_disable_other_cameras_false(self, config_rollback, camera_sample,
```

api/tests/utils/common_functions.py
```diff
@@ -13,6 +13,9 @@ def get_config_file_json_strings(config_sample_path, decamelize=False):
         config_sample_json[section] = config_sample.get_section_dict(section)
 
     if decamelize:
+        """
+        It is strange how humps.decamelize works, beacuse "Source_1" turns into "source__1". To take into account.
```
renzodgc, line 17: Yeah, this behaviour is not the most useful... 
```suggestion
        It is strange how humps.decamelize works, beacuse "Source_1" turns into "source__1".
```

api/tests/utils/common_functions.py
```diff
@@ -79,6 +82,15 @@ def get_app_from_config_file(config_sample_path):
 # Get App from config file -- end
 
 
+# Get any section from config file -- begin
+def get_section_from_config_file(section, config_sample_path):
+    config_sample_json = get_config_file_json_strings(config_sample_path)
+    config_mapped_string = map_section_from_config(section, config_sample_json)
+    config_mapped = section_string_to_section_multi_type(config_mapped_string)
+    return config_mapped
+# Get any section from config file -- end
```
renzodgc, line 91: why are these comments here?

api/tests/data/config-x86-openvino.ini
```diff
@@ -142,3 +142,32 @@ Enabled = False
 ; Expressed in minutes
 BackupInterval = 30
 BackupS3Bucket = your-s3-bucket
+
```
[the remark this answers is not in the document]
JFer11: Of course, however, it is not the original configuration file. Instead, is a copy used for testing purposes only.

api/tests/utils/common_functions.py
JFer11: That docstring was removed, and a new line was added.

api/tests/app/test_camera.py
JFer11: Thank you. And yes, it is still relevant.

api/tests/app/test_camera.py
JFer11: Yes, not necessary at all, it was just a personal convention. Should I remove them from all classes?

api/tests/app/test_camera.py
JFer11: Perfect, will be changed!

api/tests/utils/fixtures_tests.py
JFer11: The point is that, when we create a camera to test something, that camera creates a folder within the path "data / processor / static / screenshots / <camera_id>". So, once the test was done, we had to delete said directory and all its contents. This task is taken by this fixture and its tearDown code.

api/tests/utils/fixtures_tests.py
JFer11: Yes, `shutil.rmtree (path)` removes a directory and all its contents. In our case, inside the deleted folder, there will be the homography matrix and, sometimes, an object_log directory. But both directories need to be removed once the test is finished.

api/tests/utils/common_functions.py
JFer11: Personally, I did this to group functions that have a common purpose, but I can remove those comments if I want to.

api/tests/app/test_camera.py
JFer11: Yes, the comments refer to the first test only.

api/tests/app/test_camera.py
JFer11: Nice :+1:

api/tests/utils/fixtures_tests.py
mats-claassen: I mean the `yield None`

api/tests/app/test_camera.py
renzodgc: Yeah I'd remove the "Class" part

api/tests/app/test_camera.py
renzodgc: Reading this again I realised my message wasn't quite clear :sweat_smile:. Just in case of doubt I meant checking on the message string in the error as you did with other tests

api/tests/app/test_camera.py
renzodgc: I'd either compare the successful  json to what's expected or just remove the comment

api/tests/utils/common_functions.py
renzodgc: Yeah I think it adds a bit of clutter, If wanted you could add a single line docstring to each method, else the name should suffice.

api/tests/utils/fixtures_tests.py
JFer11: Pytest provide an intuitive way to run setUp and tearDown code with fixtures.
Simply put, the setup code runs before the return statement. But if we want to add tearDown code, we just have to change return -> yield, and write our tearDown code after the yield statement.
It is explained here: https://docs.pytest.org/en/stable/fixture.html#yield-fixtures-recommended

api/tests/app/test_camera.py
JFer11: Okey, changed!

api/tests/app/test_camera.py
JFer11: Okey, changed!

api/tests/app/test_camera.py
JFer11: Perfect :)
Don t worry, it was clear. I changed it, and now it checks for the correct response.json()

api/tests/utils/common_functions.py
JFer11: No problem, will be removed!

api/tests/utils/common_functions.py
```diff
@@ -13,6 +13,9 @@ def get_config_file_json_strings(config_sample_path, decamelize=False):
         config_sample_json[section] = config_sample.get_section_dict(section)
 
     if decamelize:
+        """
+        It is strange how humps.decamelize works, beacuse "Source_1" turns into "source__1". 
+        """
```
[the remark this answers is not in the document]
JFer11: Yes, however, it is strange, because documentation says that will only split on capital letters.
https://github.com/domchristie/humps#humpsdecamelizestring-options

api/tests/app/test_camera.py
JFer11: Thanks for pointing that out. I replaced it with a "create_n_cameras" function to solve this inconvenience.

api/tests/app/test_camera.py
```diff
@@ -17,7 +18,17 @@ def create_a_camera(client, camera):
...
 
 
-def create_2_cameras(client, camera_base, enable=False):
+def create_n_cameras(client, camera_base, number_of_cameras, enable=False):
+    list_of_created_cameras = []
+    for i in range(number_of_cameras):
+        camera = copy.deepcopy(camera_base)
+        del camera['id']
+        # camera['id'] = None
+        # camera['id'] = i+60
```
renzodgc, line 27: Remove these comments, please :)
```suggestion
```

JFer11: Oh yes, my fault, thanks :)

renzodgc (approved):

PR #133, opened 2021-02-09
Added PoC instructions to readme and polished configs

renzodgc:


README.md
```diff
@@ -89,6 +89,21 @@ There are two alternatives to run the processor in your device:
...
+      a. **If the device has access to an Nvidia GPU:** GPU with TensorRT optimization.
+      b. **If the device has access to an Intel CPU:** x86 using OpenVino.
+      c. **If the device has access to a CPU:** x86.
+   2. Either **build the image or pull it from Dockerhub**. Don't forget to follow the script and download the model.
+   3. Download the sample video running `./download_sample_video.sh`.
+   4. Run the processor using the script listed in its device.
+
+This way you can skip security steps such as enabling HTTPs communication or oauth and get a simple version of the processor running to see if it fits your use case.
+
+Afterwards if you intend on running the processor while consuming from a dedicated videofeed we advice you to return to this README and reat it fully.
```
pgrill, line 105: I think that we can add some `,`. Something like

```
Afterward, if you intend on running the processor while consuming from a dedicated video feed, we advise you to return to this README and read it fully.
```

pgrill (commented):
Maybe we can add a comment mentioning that they can remove the classifier step if they are not interested in face mask detections. Also, we should enable the metrics steps by default or mention in the Readme that this step is required to use the dashboard.

README.md
```diff
@@ -89,6 +89,21 @@ There are two alternatives to run the processor in your device:
   1. Using `git` and building the docker image yourself (Follow the guide in [this](#running-the-processor-building-the-image) section). 
   2. Pulling the (already built) image from [Neuralet's Docker Hub repository](https://hub.docker.com/repository/docker/neuralet/smart-social-distancing) (Follow the guide in [this](#running-the-processor-from-neuralet-docker-hub-repository) section).
 
+##### Running a proof of concept
+
+If you want to simply run the processor for just trying it out, then from the following steps you should only:
+   1. Select your device and find its image. On x86, without a dedicated edge device, you should use either:
```
mats-claassen, line 95:
```suggestion
   1. Select your device and find its docker image. On x86, without a dedicated edge device, you should use either:
```

README.md
```diff
@@ -89,6 +89,21 @@ There are two alternatives to run the processor in your device:
...
+      a. **If the device has access to an Nvidia GPU:** GPU with TensorRT optimization.
+      b. **If the device has access to an Intel CPU:** x86 using OpenVino.
+      c. **If the device has access to a CPU:** x86.
```
mats-claassen, line 98: I would just put: "Otherwise: x86". I don't think any computer wouldn't have a CPU 😄

renzodgc: :stuck_out_tongue_closed_eyes:

README.md
```diff
@@ -507,7 +507,10 @@ All the configurations are grouped in *sections* and some of them can vary depen
 
 - `[Classifier]`:
 
-  Some of the supported devices includes the *facemask detection* feature. If you want to include this feature, you need to specify the classifier section.
+  Some of the supported devices include models that allow for detecting the body-skeleton of a person.
```
mats-claassen, line 510: I believe `pose` is more common than skeleton for this task

README.md
```diff
@@ -507,7 +507,10 @@ All the configurations are grouped in *sections* and some of them can vary depen
...
+  This is a key component to **Facemask Detection**.
+  If you want to include this feature, you need to uncomment this section, and use a model that supports the Classifier.
+  Else you can delete or uncomment this section of the config file to save on CPU usage.
```
mats-claassen, line 513:
```suggestion
  Otherwise, you can delete or uncomment this section of the config file to save on CPU usage.
```

or something similar. I think `Else` is not used like in programming

mats-claassen (approved):

PR #135, opened 2021-02-12
In/Out Endpoint Configuration

renzodgc:
- Add endpoints to support the In/Out Metric.
- Add structure for the metric's implementation.
- Fixed typos in metrics interface.


api/models/camera.py
```diff
@@ -61,3 +61,16 @@ class Config:
                 'contour_roi': [[88, 58], [90, 284], [279, 284], [281, 58]]
             }
         }
+
+
+class InOutBoundaries(BaseModel):
+    in_line: Tuple[Tuple[int, int], Tuple[int, int]]
```
mats-claassen, line 67: You will also need the direction of movement for each line. Not sure about the best way to store it

renzodgc: Wasn't it from `in` to `out`?

api/routers/cameras.py
```diff
@@ -340,3 +342,47 @@ async def remove_roi_contour(camera_id: str, reboot_processor: Optional[bool] =
...
+
+
+@cameras_router.get("/{camera_id}/in_out_boundaries")
+async def get_in_out_boundaries(camera_id: str):
+    """
+        Get the In/Out Boundaries
+    """
+    in_out_file_path = InOutMetric.get_in_out_file_path(camera_id, settings.config)
+    in_out_boundaries = InOutMetric.get_in_out_boundaries(in_out_file_path)
+    if in_out_boundaries is None:
```
pgrill, line 354: I think that before checking for the boundaries we should check for camera existence in order to return a clear 404 message.

libs/metrics/in_out.py
```diff
@@ -0,0 +1,62 @@
...
+    @classmethod
+    def get_weekly_report(cls, entities: List[str], number_of_weeks: int = 0,
+                          from_date: date = None, to_date: date = None) -> Dict:
+        raise NotImplementedError("Operation not implemented")
+
+    @classmethod
+    def get_in_out_file_path(cls, camera_id, config):
+        """ Returns the path to the roi_contour file """
+        return f"{get_source_log_directory(config)}/{camera_id}/{IN_OUT}/{IN_OUT}.json"
+
```
pgrill, line 51: I am not sure if the log directory is the best place to store these configuration files (such as the calibration matrix and the in/out json).
Maybe we can "refactor" this and store the config files in a different folder.

renzodgc: Yeah, I didn't want to clog the camera's configs, and the file should be inside one of the folders for a specific camera

api/routers/cameras.py
renzodgc: Huh, I thought we were doing that by default on a callback, I'll update the endpoints to check for camera existance

api/models/camera.py
mats-claassen: Nope. Each line has a direction. Only if a person is moving in that direction it will be counted. In/Out could possibly be the same line but will have opposite directions

api/routers/cameras.py
```diff
@@ -236,6 +239,7 @@ async def get_camera_image(camera_id: str):
     """
     Gets the image related to the camera <camera_id>
     """
+    _ = get_camera_from_id(camera_id)
```
mats-claassen, line 242: While this probably works, it doesn't sound right. I think you are misusing the function or at least the name is very misleading

api/routers/cameras.py
```diff
@@ -347,8 +355,12 @@ async def remove_roi_contour(camera_id: str, reboot_processor: Optional[bool] =
 @cameras_router.get("/{camera_id}/in_out_boundaries")
 async def get_in_out_boundaries(camera_id: str):
     """
-        Get the In/Out Boundaries
+        Get the In/Out Boundaries for a camera.
+        Two coordinates `[x,y]` are given in duples `[A,B]`. These points form a **line**.
```
mats-claassen, line 359:
```suggestion
        Two coordinates `[x,y]` are given in tuples `[A,B]`. These points form a **line**.
```

mats-claassen (approved):

PR #136, opened 2021-02-16
Draw Region of Interest Contour in Live Feed

renzodgc:
Draw Region of Interest Contour in Live Feed

### DIFF
--- a/libs/loggers/source_loggers/video_logger.py
+++ b/libs/loggers/source_loggers/video_logger.py
@@ -4,7 +4,7 @@ import os
 import shutil
 
 from libs.utils import visualization_utils
-
+from libs.source_post_processors.objects_filtering import ObjectsFilteringPostProcessor
 
 class VideoLogger:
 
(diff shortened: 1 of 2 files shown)


libs/loggers/source_loggers/video_logger.py
```diff
@@ -94,6 +95,11 @@ def update(self, cv_image, objects, post_processing_data, fps):
         birds_eye_window = np.zeros(self.birds_eye_resolution[::-1] + (3,), dtype="uint8")
         class_id = int(self.config.get_section_dict('Detector')['ClassID'])
 
+        roi_contour = ObjectsFilteringPostProcessor.get_roi_contour(self.roi_file_path)
+        if roi_contour is not None:
+            color = (41, 127, 255) # #ff7f29 (255, 127, 41)
```
mats-claassen, line 100: Did you validate the color with the Design guys? I think orange looks something between yellow and red and might give the idea of an issue/warning. As this is a neutral info, maybe some gray would be good. But this is just my opinion

mats-claassen (approved):

libs/loggers/source_loggers/video_logger.py
renzodgc: It's the same colour as they use to select the area in the step before (Lanthorn's Orange). Maybe after validating it in the sprint review we can change the colour if a  change gets suggested

pgrill (approved):

PR #139, opened 2021-02-25
In Out Live-data analytics and logic

renzodgc:


libs/metrics/in_out.py
```diff
@@ -1,25 +1,58 @@
...
+                 (1, 0) - if the object entered (in)
+                 (0, 1) - if the object left (out)
+                 (0, 0) - if the object didn't cross the boundary.
+        """
+        number_of_cuts = 5
+        if len(trajectory_path) < number_of_cuts:
+            number_of_cuts = len(trajectory_path)
+
+        intersections = [trajectory_path[int(i)] for i in np.linspace(0, len(trajectory_path) - 1, number_of_cuts)]
+        intersections = zip(intersections, intersections[1:])
```
mats-claassen, line 48: Maybe we can find a better name for this variable. Maybe `trajectory_step`?

libs/metrics/in_out.py
```diff
@@ -1,25 +1,58 @@
...
+        in_out = (0, 0)
+
+        for intersection in intersections:
```
mats-claassen, line 51: This will return the last crossing of the line only. I think it is good to return only one value, unless the camera covers a big space and a person could be in the camera's space for some time and go out. In that case we need to count 1 in and 1 out.
Also if a person comes from outside and shortly goes over the line and back you will count this a 0 in and 1 out with this code.

Ideally I think we would mix the time dimension in and say that the person needs to be "inside" for X seconds before going out to be counted if it goes in and out in the same "session". Same applies in the opposite direction

libs/metrics/in_out.py
```diff
@@ -1,25 +1,58 @@
...
+            in, out) : tuple
+                 (1, 0) - if the object entered (in)
+                 (0, 1) - if the object left (out)
+                 (0, 0) - if the object didn't cross the boundary.
+        """
+        number_of_cuts = 5
+        if len(trajectory_path) < number_of_cuts:
+            number_of_cuts = len(trajectory_path)
+
+        intersections = [trajectory_path[int(i)] for i in np.linspace(0, len(trajectory_path) - 1, number_of_cuts)]
```
mats-claassen, line 47: Ideally I would use some kind of outlier removal instead of taking every Nth step but maybe that is for a future version

libs/loggers/source_loggers/raw_data_logger.py
```diff
@@ -63,6 +63,7 @@ def format_objects(self, objects_list):
             obj = {}
             # TODO: Get 3D position of objects
             obj["position"] = [0.0, 0.0, 0.0]
+            obj["bbox_real"] = obj_dict["bboxReal"]
```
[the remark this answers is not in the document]
renzodgc: It is the pixel coordinates (as the line is also marked in pixels).

I thought about using the position that is marked as 0,0,0. Would that make sense for using the middle of the bottom of the bounding box?

libs/metrics/base.py
```diff
@@ -194,7 +196,11 @@ def compute_live_metrics(cls, config, live_interval):
                 return
             entity["base_directory"] = entity_directory
             entries_in_interval = int(live_interval * 60 / get_source_logging_interval(config))
-            live_data = cls.generate_live_csv_data(today_entity_csv, entity, entries_in_interval)
+            if cls.metric_name == IN_OUT:
+                boundary_line = cls.get_in_out_boundaries(cls.get_in_out_file_path(entity["id"], config))
```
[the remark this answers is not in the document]
renzodgc: The issue is that in the metric the function has been called and I am raising an error. Here I don't want to call the metric and rise a CSV. But yesterday I thought of a better way of doing so, will update :)

libs/metrics/in_out.py
```diff
@@ -1,25 +1,58 @@
...
+            Two coordinates [x,y] are in 2-tuples [A,B]
+                Boundaries of the in/out line.
+                If someone crosses the line while having A to their right, they are going in the in direction (entering)
+                Crossing the line while having A to their left means they are going in the out direction (leaving)
+
+            trajectory_path: List of N 2-tuples (x,y)
+            That represents the trajectory of an object.
+
+        Returns:
+            in, out) : tuple
```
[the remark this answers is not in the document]
renzodgc: My idea was to have it return two integers (in, out), although it's always 1,0, 0,1 or 0,0. The issue with -1,0,1 is that I'd have to add ifs to parse it but we can change it.

libs/metrics/in_out.py
renzodgc: Yeah, I discussed it with Pablo and this seemed to be the better option

libs/metrics/in_out.py
renzodgc: So you'd rather count every intersection with the line and return the sum of in_out rather than having policyy for selecting one?

mats-claassen: No. We should only return 1. Or maybe 1 for in and 1 for out (this if we check the time a user spents in or out, likely not v1)

But with the current code if a user goes in, and before leaving the screen goes out you are counting 0 in and 1 out which is not correct

libs/loggers/source_loggers/raw_data_logger.py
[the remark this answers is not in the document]
mats-claassen: Using position would make sense if we used world coordinates. In that case using the middle of the bottom line would be fine. But I don't think we should use it with pixel coordinates

libs/metrics/in_out.py
renzodgc: Ah, I see what you mean. So then we should return (1,1), right?

My issue there is that if the user goes in, out, and then in again it should be (1,0) or (2,1).

Maybe we could return (1,0) by normalizing (2,1) to (1,0), would  that make  sense?

libs/loggers/source_loggers/video_logger.py
```diff
@@ -2,9 +2,13 @@
 import numpy as np
 import os
 import shutil
+import logging
 
 from libs.utils import visualization_utils
 from libs.source_post_processors.objects_filtering import ObjectsFilteringPostProcessor
+from libs.metrics.in_out import InOutMetric
```
pgrill, line 9: I think that you can delete this import

libs/metrics/base.py
```diff
@@ -26,6 +26,7 @@ def parse_date_range(dates):
 
 
 class BaseMetric:
+    metric_name = None
```
pgrill, line 29: Why this attribute was added?  I don't find any usage of it

libs/metrics/base.py
```diff
@@ -194,7 +195,9 @@ def compute_live_metrics(cls, config, live_interval):
                 return
             entity["base_directory"] = entity_directory
             entries_in_interval = int(live_interval * 60 / get_source_logging_interval(config))
-            live_data = cls.generate_live_csv_data(today_entity_csv, entity, entries_in_interval)
+            if not cls.can_execute(config, entity):
```
pgrill, line 198: This condition, should be also checked  in the hourly, daily and weekly methods?

libs/metrics/in_out.py
```diff
@@ -1,25 +1,57 @@
...
+                 (1, 0) - if the object entered (in)
+                 (0, 1) - if the object left (out)
+                 (0, 0) - if the object didn't cross the boundary.
+        """
+        number_of_cuts = 5
```
pgrill, line 42: Maybe we can declare number_of_cuts as a constant. What do you think?

libs/metrics/social_distancing.py
```diff
@@ -10,11 +10,12 @@
 from libs.utils.loggers import get_source_log_directory
 
 from .base import BaseMetric
+from constants import SOCIAL_DISTANCING
 
 
 class SocialDistancingMetric(BaseMetric):
 
-    reports_folder = "social-distancing"
+    reports_folder = SOCIAL_DISTANCING
```
pgrill, line 18: Should we also use the constants in the facemask metrics?

libs/metrics/base.py
renzodgc: Forgot to delete it here

libs/metrics/in_out.py
renzodgc: In the class you mean? I don't feel use on putting it on constants.py but yeah I can make it a constant in the class

libs/metrics/social_distancing.py
renzodgc: The folder for facemask is `face-mask-usage`, where as the c constant is `facemask-usage`. Chaning one would break some cases data is already stored there. We could maybe internally use `face-mask-usage` and change the constant but maybe it's best doing so in a separate PR

libs/loggers/source_loggers/raw_data_logger.py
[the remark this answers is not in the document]
renzodgc: So using the bbox_real is what makes sense then?

libs/metrics/base.py
renzodgc: I presume yes, my intent was on doing those on a separate PR an see but I could add them in this very same PR

libs/metrics/in_out.py
mats-claassen: I think that the correctness of changing (2,1) to (1,0) depends on the time the person spends in and out. Like if the bbox detection flickers a bit and jumps above and below the line then we don't want to count that as multiple crosses. But if the user spends some time then maybe yes. We could also define this as a minimum of N frames on one side before we count a user that goes back and forth as (1,1). If it is les than N frames we take it as (0,0)

config-x86-gpu-tensorrt.ini
```diff
@@ -25,7 +25,7 @@ SlackChannel = lanthorn-notifications
 OccupancyAlertsMinInterval = 180
 MaxThreadRestarts = 5
 GlobalReportingEmails =
-GlobalReportTime = 06:00
+GlobalReportTime = 14:42
```
mats-claassen, line 28: I guess this change and the ones below are not on purpose

libs/metrics/in_out.py
```diff
@@ -15,11 +16,31 @@
 from libs.utils.config import get_source_config_directory
 from libs.utils.utils import validate_file_exists_and_is_not_empty
 
+logger = logging.getLogger(__name__)
```
mats-claassen, line 19: Doesn't seem to be used

libs/metrics/in_out.py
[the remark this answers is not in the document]
pgrill: As Mathi mentioned, I think that line 59 should be `(in, out) :  tuple`

libs/metrics/in_out.py
```diff
@@ -1,61 +1,263 @@
...
+                    paths[track_id].extend(sub_path)
+        for track_id, data in paths.items():
+            path = data["path"]
+            new_in, new_out = cls.process_path(boundary_line, path)
+            people_in += new_in
+            people_out += new_out
+        return [people_in, people_out]
 
     @classmethod
     def get_trend_live_values(cls, live_report_paths: Iterator[str]) -> Iterator[int]:
```
pgrill, line 136: Are we showing the indicator trend in the dashboard? I am not sure if makes sense to calculate it if is not used in any place.

libs/metrics/in_out.py
```diff
@@ -1,61 +1,263 @@
...
+    @classmethod
+    def can_execute(cls, config, entity):
+        boundary_line = cls.read_in_out_boundaries(cls.get_in_out_file_path(entity["id"], config))
+        if boundary_line is None:
+            return False
+        return True
+
+# Auxiliary methods taken from:
+# https://github.com/yas-sim/object-tracking-line-crossing-area-intrusion/blob/master/object-detection-and-line-cross.py
+def check_line_cross(boundary_line, trajectory):
```
pgrill, line 182: Maybe we can move these functions to a `utils` file. What do you think?

libs/metrics/in_out.py
renzodgc: I don't think so, I don't think it hurts too much do a "basic version" just to simplify its integration

libs/metrics/in_out.py
renzodgc: Hmm I thought about reducing complexity here, although it will be used only in in/out. I'd maybe leave it here until the next iteration (multiple lines, areas, etc)

libs/loggers/source_loggers/video_logger.py
```diff
@@ -2,10 +2,13 @@
 import numpy as np
 import os
 import shutil
+import logging
 
 from libs.utils import visualization_utils
 from libs.source_post_processors.objects_filtering import ObjectsFilteringPostProcessor
 
+logger = logging.getLogger(__name__)
```
mats-claassen, line 10: This does not look as if it is used

libs/metrics/base.py
```diff
@@ -338,3 +344,7 @@ def get_live_report(cls, entities):
         if trend_live_values:
             report["Trend"] = cls.calculate_trend_value(trend_live_values)
         return report
+
+    @classmethod
```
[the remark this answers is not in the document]
renzodgc: Done

PR #147, opened 2021-03-18
Config multiple In/Out per camera

renzodgc:


libs/metrics/in_out.py
```diff
@@ -168,87 +177,41 @@ def can_execute(cls, config, entity):
...
+                weekly_out = []
+            else:
+                boundary_names, weekly_in, weekly_out = list(zip(*week_data["Summary"]))
+                boundary_name = next(x for x in boundary_names if not is_list_recursively_empty(x))
+                weekly_in = fill_partially_empty_result(weekly_in, 0)
+                weekly_out = fill_partially_empty_result(weekly_out, 0)
+
+            report["Summary"].append([
+                boundary_name,
+                [sum(x) for x in zip(*weekly_in)],
```
mats-claassen, line 205: It is hard to follow this code without info about types of these variables. Is weekly_in not an array of numbers where you could directly do sum(weekly_in)?

mats-claassen (approved):

pgrill (approved):

PR #148, opened 2021-03-19
Different models per camera

JFer11:
Before, all cameras run the same ML model to process video. Now each camera supports its own ML model and while one camera can run a model A, another camera can run a model B.


libs/detectors/x86/detector.py
```diff
@@ -1,3 +1,37 @@
...
+import os
+from pathlib import Path
+
+
+def get_or_create_model_json_file(config, device, camera_id):
+    """
+    If no model directory was created for a given device, the default values will be taken from the configuration file
+    and a model_<device>.json file will be created.
+    """
+    base_path = os.path.join(config.get_section_dict("App")["EntityConfigDirectory"], "sources", camera_id)
```
renzodgc, line 11: Please use the helper defined in libs/utils/config.py

libs/detectors/x86/detector.py
```diff
@@ -1,3 +1,37 @@
...
+    models_directory_path = os.path.join(base_path, "ml_models")
+    json_file_path = os.path.join(models_directory_path, f"model_{device}.json")
+
+    if not os.path.exists(json_file_path):
+        # Hypothesis: source config directory (base_path) should always exists.
+        if not os.path.exists(models_directory_path):
+            Path(models_directory_path).mkdir(parents=True, exist_ok=True)
```
renzodgc, line 18: I think we can run this line without the ifs (as it's okay if it already exists and creates the parent's folders)

libs/detectors/x86/detector.py
```diff
@@ -1,3 +1,37 @@
...
+        # Create .json file
+        json_content = {
+            "model_name": config.get_section_dict("Detector")["Name"],
+            "variables": {
+                key: value for key, value in config.get_section_dict("Detector").items() if key not in ["Name", "Device"]
+            }
+        }
+        with open(json_file_path, 'x+') as outfile:
+            json.dump(json_content, outfile)
+        os.chmod(json_file_path, 0o777)
```
renzodgc, line 29: Why is this? Because access is only alowed to the docker user?

model_x86.json
```diff
@@ -0,0 +1 @@
+{"model_name": "SABORR", "variables": {"ImageSize": "300,300,3", "ModelPath": "", "ClassID": "1", "MinScore": "0.25"}}
```
renzodgc, line 1: IS this in root? The ml_folders should be included in the gitignore

libs/detectors/x86/detector.py
```diff
@@ -7,18 +41,23 @@ class Detector:
...
+
+        self.name = model_data["model_name"]
 
         if self.name == 'mobilenet_ssd_v2':
             from libs.detectors.x86 import mobilenet_ssd
             self.net = mobilenet_ssd.Detector(self.config)
         elif self.name == "openvino":
             from libs.detectors.x86 import openvino
-            self.net = openvino.Detector(self.config)
+            self.net = openvino.Detector(self.config, self.name, model_data["variables"])
```
renzodgc, line 60: This change must be made for all detectors and their models.

libs/detectors/x86/detector.py
```diff
@@ -1,3 +1,37 @@
...
+        if not os.path.exists(models_directory_path):
+            Path(models_directory_path).mkdir(parents=True, exist_ok=True)
+
+        # Create .json file
+        json_content = {
+            "model_name": config.get_section_dict("Detector")["Name"],
+            "variables": {
+                key: value for key, value in config.get_section_dict("Detector").items() if key not in ["Name", "Device"]
+            }
+        }
```
renzodgc, line 26: Do these inherit from a generic detector.py? Should this be done there?

libs/detectors/x86/detector.py
mats-claassen: You definitely don't need execute permissions on a json file

libs/detectors/x86/detector.py
mats-claassen: You would still need the first if so I think it is okay this way. It would work without ifs but this is clearer I think

libs/detectors/x86/detector.py
```diff
@@ -7,18 +41,23 @@ class Detector:
     input image in order to get the detection results.
 
     :param config: Is a ConfigEngine instance which provides necessary parameters.
+    :param source: A string that represents the camera the camera. Ex: "Source_1".
```
mats-claassen, line 44:
```suggestion
    :param source: A string that represents the camera. Ex: "Source_1".
```

model_x86.json
mats-claassen: These files should be stored in some subfolder of the `data` folder (on that volume)

libs/detectors/x86/detector.py
```diff
@@ -1,3 +1,37 @@
+import json
+import os
+from pathlib import Path
+
+
+def get_or_create_model_json_file(config, device, camera_id):
```
mats-claassen, line 6: We will need to update these json files in the future with the latest version of the config file. So I think it would be good to split the creation part into another function, which does overwrite the json file if it already exists

libs/detectors/x86/detector.py
```diff
@@ -1,3 +1,37 @@
...
+    json_file_path = os.path.join(models_directory_path, f"model_{device}.json")
+
+    if not os.path.exists(json_file_path):
+        # Hypothesis: source config directory (base_path) should always exists.
+        if not os.path.exists(models_directory_path):
+            Path(models_directory_path).mkdir(parents=True, exist_ok=True)
+
+        # Create .json file
+        json_content = {
+            "model_name": config.get_section_dict("Detector")["Name"],
```
mats-claassen, line 22: This does not allow using different types of detectors per camera. it will take the same Detector parameters for each camera. We need to specify different Detector parameters per camera (and have defaults). So this requires some structure changes in the config files.
The idea is to use this capability to use adapted models for one camera. So for that camera we need to know all the parameters that are currently in the Detector section. However some of them are constant for all models and others will be the same for each adapted camera. The only thing that will be different for each adapted camera is the ModelPath. (But some other parameters will be different from the default values)

If this is confusing, we can chat about it

mats-claassen:
One more comment, I see lots of merged commits in here. Try to create branches of new features on master or rebase onto master

libs/detectors/detector.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, config):
             self.detector = DummyDetector(self.config)
         elif self.device in ["x86", "x86-gpu"]:
             from libs.detectors.x86.detector import Detector as X86Detector
-            self.detector = X86Detector(self.config)
+            self.detector = X86Detector(self.config, source)
```
pgrill, line 32: I think that we need to send the source to all the constructors, not only the X86

libs/detectors/x86/detector.py
pgrill: Agree

libs/detectors/x86/detector.py
```diff
@@ -1,3 +1,37 @@
...
+
+def get_or_create_model_json_file(config, device, camera_id):
+    """
+    If no model directory was created for a given device, the default values will be taken from the configuration file
+    and a model_<device>.json file will be created.
+    """
+    base_path = os.path.join(config.get_section_dict("App")["EntityConfigDirectory"], "sources", camera_id)
+    models_directory_path = os.path.join(base_path, "ml_models")
+    json_file_path = os.path.join(models_directory_path, f"model_{device}.json")
+
```
pgrill, line 14: The location and name of the custom models should be documented in the README.

libs/detectors/x86/detector.py
JFer11: Changed!

libs/detectors/x86/detector.py
JFer11: Personally, I think this will save a lot of effort if a problem ever arises.

model_x86.json
JFer11: As it is part of the config, I decided to include it in said directory: data/processor/config/sources/0/ml_models/model_x86.json.
Should I change this path? Which one would you recommend?

libs/detectors/x86/detector.py
JFer11: Of course, this commit was just a proof of concept in order to validate these changes. I will generalize it.

libs/detectors/x86/detector.py
JFer11: No, I was here just to validate the idea. In the next commit I will generalize this behavior to all `Detectors` and this function will be in another file.

libs/detectors/x86/detector.py
JFer11: Changed!

libs/detectors/detector.py
JFer11: Of course, this commit was just a proof of concept in order to validate these changes. I will generalize it.

libs/detectors/x86/detector.py
JFer11: Yes, I believe there are several issues we should tackle if we set the model in the config file, instead of a separate file.
So, first of all, we decided to automatically create the file and, if necessary, modify it manually. However, maybe we can create an endpoint to modify the model of a camera.

libs/detectors/x86/detector.py
mats-claassen: rw access for everyone maybe?

model_x86.json
mats-claassen: That path seems good. Just this file seems to be in the root directory

libs/detectors/x86/detector.py
renzodgc: But this makes sense, right? it's only changing the model path either manually or through and endpoint

libs/detectors/x86/detector.py
JFer11: Great!

libs/detectors/x86/detector.py
JFer11: This function will be modified as we discuss in chat. :)

libs/detectors/x86/detector.py
JFer11: Done!

api/models/ml_model.py
```diff
@@ -0,0 +1,133 @@
...
+    def validate_device(cls, device):
+        if device not in ["x86", "x86-gpu", "Jetson", "EdgeTPU", "Dummy"]:
+            raise ValueError('Not valid Device. Try one of the following: "x86", "x86-gpu", "Jetson", "EdgeTPU",'
+                             '"Dummy".')
+        return device
+
+    @validator("tensorrt_precision")
+    def validate_tensorrt_precision(cls, tensorrt_precision):
+        if tensorrt_precision is not None:
+            if tensorrt_precision not in [16, 32]:
```
mats-claassen, line 26: You could combine the two ifs

api/models/ml_model.py
```diff
@@ -0,0 +1,133 @@
...
+                raise ValueError('Not valid tensorrt_precision. Accepted values: 16, 32')
+
+        return tensorrt_precision
+
+    @validator("name")
+    def validate_name(cls, name):
+        if name not in ["openvino", "openpifpaf_tensorrt", "mobilenet_ssd_v2", "openpifpaf", "yolov3",
```
mats-claassen, line 33: This validation is conditional to the device type. Not every one of these is supported for all devices. But I don't know if we want to maintain such a validation. Opinions?

api/models/ml_model.py
```diff
@@ -0,0 +1,133 @@
...
+    @validator("image_size")
+    def validate_image_size(cls, image_size):
+        integers = image_size.split(",")
+
+        if len(integers) != 3:
+            raise ValueError('ImageSize must be a string with 3 numbers separated with commas. Ex: "30,30,40".')
+
+        for x in integers:
+            try:
+                int(x)
```
mats-claassen, line 51: You could use list comprehension here

api/models/ml_model.py
mats-claassen: Oh I see you added that separately. What is the point of this validator given the `check_models_and_device` function?

api/routers/ml_models.py
```diff
@@ -0,0 +1,66 @@
...
+    if "ClassId" in result.keys():
+        result["ClassID"] = result["ClassId"]
+        del result["ClassId"]
+
+    return result
+
+
+@ml_model_router.post("/{camera_id}")
+async def modify_ml_model(camera_id: str, model_parameters: MLModelDTO):
+    # TODO: What about REBOOT PROCESOR? Is it necessary here?
```
mats-claassen, line 35: Yes I think so

api/tests/app/test_ml_models.py
```diff
@@ -0,0 +1,397 @@
...
+import pytest
+import os
+import json
+
+# The line below is absolutely necessary. Fixtures are passed as arguments to test functions.
+# This is why the IDE cannot recognize them.
+from api.tests.utils.fixtures_tests import config_rollback_cameras, rollback_camera_config
+
+
+def pascal_case_to_snake_case(parameters):
```
mats-claassen, line 12: Is this function not the same as the one defined in the app? Why not define it in utils and import in both?

libs/detectors/detector.py
```diff
@@ -11,7 +11,10 @@ def __init__(self, config, source):
         self.config = config
         self.device = self.config.get_section_dict("Detector")["Device"]
         self.resolution = tuple([int(i) for i in self.config.get_section_dict("App")["Resolution"].split(",")])
+        # TODO: Iterator in the below line should be replaced with:
```
mats-claassen, line 14: Why this? Can we solve it?

libs/detectors/utils/ml_model_functions.py
```diff
@@ -0,0 +1,33 @@
...
+    Check if the json file exists and return the information about the ML model within said file. In case the file does
+    not exist, return defaults values from "Detector" in config file.
+    """
+    base_path = os.path.join(get_source_config_directory(config), camera_id)
+    models_directory_path = os.path.join(base_path, "ml_models")
+    json_file_path = os.path.join(models_directory_path, f"model_{device}.json")
+
+    # Hypothesis: source config directory (base_path) should always exists.
+
+    if not os.path.exists(models_directory_path):
```
mats-claassen, line 18: I don't think this folder should be created here. This is a "get" function and I don't think we gain anything by creating the folder here, right?

libs/detectors/x86/openpifpaf.py
```diff
@@ -12,17 +12,20 @@ class Detector:
...
-        self.device = "cuda:0" if self.config.get_section_dict('Detector')['Device'] == "x86-gpu" else "cpu"
+        self.model_name = model_name
+        self.model_variables = variables
+        self.device = "cuda:0" if self.config.get_section_dict("Detector")["Device"] == "x86-gpu" else "cpu"
         # Frames Per Second
         self.fps = None
         self.net, self.processor = self.load_model()
-        self.w, self.h, _ = [int(i) for i in self.config.get_section_dict('Detector')['ImageSize'].split(',')]
+        self.w, self.h, _ = [int(i) for i in self.model_variables['ImageSize'].split(',')]
+        # self.w, self.h, _ = [int(i) for i in self.config.get_section_dict('Detector')['ImageSize'].split(',')]
```
mats-claassen, line 28: Please remove commented lines

README.md
```diff
@@ -585,6 +594,7 @@ The available endpoints are grouped in the following subapis:
 - `/export`: an endpoint to export (in zip format) all the data generated by the processor.
 - `/slack`: a set of endpoints required to configure Slack correctly in the processor. We recommend to use these endpoints from the [UI](https://app.lanthorn.ai) instead of calling them directly.
 - `/auth`: a set of endpoints required to configure OAuth2 in the processors' endpoints.
+- `/ml_model`: an endpoint to edit the ML model and the model parameters that a camera will use to process its video.
```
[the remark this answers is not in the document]
JFer11: Done!

api/models/ml_model.py
JFer11: Changed!

api/models/ml_model.py
JFer11: The truth is that it is enough for validations if I remove `validate_name`. However, this validator will help the user to know what all the options are available for `model_name`.

api/models/ml_model.py
JFer11: Great!

api/routers/ml_models.py
JFer11: I added that option.

api/models/ml_model.py
```diff
@@ -0,0 +1,133 @@
...
+                raise ValueError(f'The device {values.get("device")} only supports the following models:'
+                                 f'"mobilenet_ssd_v2", "openvino","openpifpaf","openpifpaf_tensorrt"'
+                                 f'"yolov3". ')
+
+        return values
+
+    @root_validator(skip_on_failure=True)
+    def check_variables_for_models(cls, values):
+
+        if values.get("name") == "ssd_mobilenet_v2_coco" or values.get("name") == "ssd_mobilenet_v2_pedestrian_softbio":
```
pgrill, line 88: Maybe we can simplify all the `pass` sentences in a single line:

Something like 

```
if values.get("name) in ["ssd_mobilenet_v2_coco",'"ssd_mobilenet_v2_pedestrian_softbio", ... ]:
    pass
```

api/models/ml_model.py
renzodgc: For me there isn't much, it duplicates a validation.

api/models/ml_model.py
```diff
@@ -0,0 +1,133 @@
...
+                int(x)
+            except ValueError:
+                raise ValueError('ImageSize must be a string with 3 numbers separated with commas. Ex: "30,30,40".')
+
+        return image_size
+
+    # Root validators are called after each field validators success.
+
+    @root_validator(skip_on_failure=True)
+    def check_models_and_device(cls, values):
```
renzodgc, line 60: Please add all these strings in a dictionary as a constant

MODELS_DEVICES = { device: [models], device: [models]}

And refactor the validations to reuse information from there (so it is centralized). For example: in order to get all devices MODELS_DEVICES.keys()

We could do something similar with the variables in the validator below (by instead of storing an array of models we store a dictionary of variables, models being one of them), although maybe that's too complicated for images sizes that  are a multiple of something.

api/tests/app/test_ml_models.py
renzodgc: But if it is the same implementation I see no value in copying it instead of importing it, duplication of code is even more prone to bugs, right?

renzodgc: If we want to be protected against these bugs an unitary test of  the method pascal_case_to_snake_case should be implemented, but I don't see that much value in that

README.md
```diff
@@ -564,6 +564,15 @@ All the configurations are grouped in *sections* and some of them can vary depen
       - `BackupInterval`: Expressed in minutes. Defines the time interval desired to back up the raw data.
       - `BackupS3Bucket`: Configures the S3 Bucket used to store the backups.
 
+
+#### Use different models per camera
+Before processing video from a camera, a ML model could be specified to process that camera.
+This information is stored in a .json file in the path: /repo/data/processor/config/sources/<camera_id>/ml_models/model_<device>.json
+To modify the model used, use the endpoint `/ml_model` documented below.
+This custom model must specify: "model_name" and "variables" for the model. 
+If no custom model is detected, the default values will be applied and taken from the `[Detector]` section, mentioned above.
```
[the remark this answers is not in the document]
JFer11: Beautiful! Thanks a lot!

README.md
```diff
@@ -564,6 +564,15 @@ All the configurations are grouped in *sections* and some of them can vary depen
       - `BackupInterval`: Expressed in minutes. Defines the time interval desired to back up the raw data.
       - `BackupS3Bucket`: Configures the S3 Bucket used to store the backups.
 
+
+#### Use different models per camera
+Before processing video from a camera, a ML model could be specified to process that camera.
```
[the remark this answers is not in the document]
JFer11: I changed it with the recommendations Renzo gave me

README.md
```diff
@@ -564,6 +564,15 @@ All the configurations are grouped in *sections* and some of them can vary depen
...
+This information is stored in a .json file in the path: /repo/data/processor/config/sources/<camera_id>/ml_models/model_<device>.json
+To modify the model used, use the endpoint `/ml_model` documented below.
+This custom model must specify: "model_name" and "variables" for the model.
```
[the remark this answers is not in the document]
JFer11: I changed it with the recommendations Renzo gave me

api/tests/app/test_ml_models.py
JFer11: Okey, no problem I will import that function.

JFer11: Actually, I realize that when I create the function, it was because I need the same functionality as pascal_case_to_snake_case defined in ml_models.py, but instead of giving an Object as an argument, it is necessary to provide a dictation. The function defined in test_ml_models.py differs in the for statement, the iterator is different.

JFer11: So we need to keep the function

libs/detectors/detector.py
JFer11: It was a reminder for myself, we should discuss this personally.

libs/detectors/utils/ml_model_functions.py
JFer11: No, the endpoint is responsible for creating the folder. Removed

libs/detectors/x86/openpifpaf.py
JFer11: Done!

api/models/ml_model.py
JFer11: Okey, no problem.

JFer11: Done!

api/models/ml_model.py
JFer11: Removed!

api/models/ml_model.py
JFer11: Changed!

api/models/ml_model.py
```diff
@@ -4,6 +4,18 @@
...
 
+MODELS_DEVICES = {
+    "Jetson": ["ssd_mobilenet_v2_coco", "ssd_mobilenet_v2_pedestrian_softbio", "openpifpaf_tensorrt"],
+    "EdgeTPU": ["mobilenet_ssd_v2", "pedestrian_ssd_mobilenet_v2", "pedestrian_ssdlite_mobilenet_v2", "posenet"],
+    "Dummy": ["openvino", "openpifpaf_tensorrt", "mobilenet_ssd_v2", "openpifpaf", "yolov3", "ssd_mobilenet_v2_coco",
+              "ssd_mobilenet_v2_pedestrian_softbio", "posenet", "pedestrian_ssd_mobilenet_v2",
+              "pedestrian_ssdlite_mobilenet_v2"],  # All available models.
+    "x86": ["mobilenet_ssd_v2", "openvino", "openpifpaf", "openpifpaf_tensorrt", "yolov3"],
+    "x86-gpu": ["mobilenet_ssd_v2", "openvino", "openpifpaf", "openpifpaf_tensorrt", "yolov3"],
+
```
renzodgc, line 15: why the extra line?

api/models/ml_model.py
```diff
@@ -59,23 +58,35 @@ def validate_image_size(cls, image_size):
     @root_validator(skip_on_failure=True)
     def check_models_and_device(cls, values):
         if values.get("device") == "Jetson":
-            if values.get("name") not in ["ssd_mobilenet_v2_coco", "ssd_mobilenet_v2_pedestrian_softbio",
-                                          "openpifpaf_tensorrt"]:
+            if values.get("name") not in MODELS_DEVICES["Jetson"]:
                 raise ValueError(f'The device {values.get("device")} only supports the following models:'
                                  f'"ssd_mobilenet_v2_coco", "ssd_mobilenet_v2_pedestrian_softbio",'
                                  f'"openpifpaf_tensorrt". ')
```
renzodgc: Here you should print the `MODELS_DEVICES["Jetson"]` list, same in all cases

libs/detectors/detector.py
```diff
@@ -11,11 +13,10 @@ def __init__(self, config, source):
...
         self.resolution = tuple([int(i) for i in self.config.get_section_dict("App")["Resolution"].split(",")])
-        # TODO: Iterator in the below line should be replaced with:
-        #  get_model_json_file_or_return_default_values(self.config, self.device, self.config.get_section_dict(source)["Id"])["variables"]["ImageSize"]
-        self.image_size = [int(i) for i in self.config.get_section_dict("Detector")["ImageSize"].split(",")]
-        # Do: Ctrl + Shift + F and type: "ImageSize" and "get_section_dict("Detector")" (also with this '' quotes)
-
+        self.image_size = [int(i) for i in get_model_json_file_or_return_default_values(self.config, self.device, self.config.get_section_dict(source)["Id"])["variables"]["ImageSize"].split(",")]
+        logger.info("-----------------------------------------")
+        logger.info(source)
+        logger.info("-----------------------------------------")
```
renzodgc, line 19: will these stay? If so I'd remove logging "-----" and adding a message before the  source info

libs/loggers/source_loggers/video_logger.py
```diff
@@ -93,7 +94,7 @@ def update(self, cv_image, objects, post_processing_data, fps):
         dist_threshold = post_processing_data.get("dist_threshold", 0)
 
         birds_eye_window = np.zeros(self.birds_eye_resolution[::-1] + (3,), dtype="uint8")
-        class_id = int(self.config.get_section_dict('Detector')['ClassID'])
+        class_id = int(self.config.get_section_dict('Detector')['ClassID'])  # TODO: Have a look.
```
renzodgc, line 97: Have a look regarding what?

api/models/ml_model.py
JFer11: It is not extra, it is another device "x86" and  "x86-gpu".

renzodgc: nono I meant the empty line before the }

libs/detectors/detector.py
JFer11: No, should be deleted.

libs/loggers/source_loggers/video_logger.py
JFer11: It was modified, as well as self.image_size in file libs/detectors/detector.py (line 16 and 17), self.model_input_size in libs/classifiers/jetson/face_mask.py (lines: 60-66). We should discuss what to do in endpoint get_processor_info().

api/models/ml_model.py
JFer11: Oh, sorry. Done!

api/models/ml_model.py
JFer11: Changed!

api/routers/config.py
```diff
@@ -98,6 +98,7 @@ async def update_config_file(config: ConfigDTO, reboot_processor: Optional[bool]
 
 @config_router.get("/info", response_model=ConfigInfo)
 async def get_processor_info():
+    # Here we have to specify the camera right? Or are we asking for the default values?
```
renzodgc, line 101: This part only cares about the device name, so it is a default value. It asks for the model name but just to recognize openvino.

JFer11: It is still coherent that the endpoint asks if the default model name is "openvino"?

renzodgc: Yes. As the only  model name available for openvino is "openvino". The filepath my change but it is still openvino

JFer11: Okey, perfect!

renzodgc (approved):

api/models/ml_model.py
```diff
@@ -59,49 +57,51 @@ def validate_image_size(cls, image_size):
...
-            pass
-
-        elif values.get("name") == "pedestrian_ssd_mobilenet_v2":
-            pass
-
-        elif values.get("name") == "pedestrian_ssdlite_mobilenet_v2":
-            pass
+            integers = values.get("image_size").split(",")
+            if int(integers[0]) % 16 != 1 or int(integers[0]) % 16 != 1:
+                raise ValueError('First two values of ImageSize must be multiples of 16 plus 1 for openpifpaf_tensorrt.'
```
mats-claassen, line 103: I think there are only 3 valid options but I don't think it is too important to have the exact validation here

libs/classifiers/jetson/face_mask.py
```diff
@@ -1,20 +1,23 @@
...
-import sys 
+import sys
 import time
 import logging
 import os
 from libs.detectors.utils.fps_calculator import convert_infr_time_to_fps
+from libs.detectors.utils.ml_model_functions import get_model_json_file_or_return_default_values
+
+# logging.getLogger().setLevel(logging.INFO)
+logger = logging.getLogger(__name__)
```
mats-claassen, line 12: I think the logger isn't being used in this file so these lines could be deleted

api/models/ml_model.py
JFer11: Do you suggest removing all validations related to image size?

mats-claassen (approved):

PR #150, opened 2021-03-19
Area all unique

JFer11:
Added a new unique area with id "ALL" representing all areas. This area can be used on endpoints and is not in the configuration file (.ini).


api/routers/areas.py
```diff
@@ -36,6 +36,13 @@ async def get_area(area_id: str):
     """
     Returns the configuration related to the area <area_id>
     """
+    if area_id.upper() == "ALL":
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
```
pgrill, line 41: Why the `get` operation is forbidden in the area "ALL"?
I thought that we will allow all the read operations over areas :thinking:

api/routers/areas.py
```diff
@@ -36,6 +36,13 @@ async def get_area(area_id: str):
     """
     Returns the configuration related to the area <area_id>
     """
+    if area_id.upper() == "ALL":
```
pgrill, line 39: Maybe we can create a constant called ALL_AREAS with the value "ALL"  and use It instead of repeat the string,

api/routers/metrics/metrics.py
```diff
@@ -27,7 +27,7 @@ def validate_camera_existence(camera_id: str):
 
 
 def get_areas(areas: str) -> Iterator[str]:
-    if areas:
+    if areas and areas.upper() != "ALL":
```
pgrill, line 30: `areas` is a list of strings. I think that the condition should be
```
if areas and not "ALL" in areas:
```

api/routers/areas.py
renzodgc: This operation should be valid, and return an area with all cameras

api/routers/areas.py
```diff
@@ -55,6 +62,11 @@ async def create_area(new_area: AreaConfigDTO, reboot_processor: Optional[bool]
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=bad_request_serializer("Area already exists", error_type="config duplicated area")
         )
+    elif new_area.id.upper() == "ALL":
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=bad_request_serializer("Area with ID: 'ALL' is not valid.", error_type="Invalid ID")
```
renzodgc, line 68: Here and in other places we would use string interpolation with the constant Pablo mentioned

api/routers/metrics/metrics.py
renzodgc: I don't understand this, if the constant is in the list it should be split as normal, right?

api/routers/areas.py
JFer11: Good idea!

api/routers/metrics/metrics.py
JFer11: Changed!

api/routers/areas.py
JFer11: But this will not validate the `response_model` set in:
`@areas_router.get("/{area_id}", response_model=AreaConfigDTO)`
Pydantic expects to return only one area, and if we give the ID: "All" a ValidationError from Pydantic is raised.

That was the reason why I think a good alternative was to inform the user that there is an endpoint to list all the cameras.
"Area with ID: 'ALL' is not valid. Instead, try the endpoint GET /areas to get all areas."

mats-claassen: Hmm, but does ALL refer to "one area with all cameras" or "all areas"? I think it is the first. Which is not necessary the same as the second

renzodgc: ALL refers to "one area with all cameras". So if you do GET /areas/ALL you should get the hardcodded information of an inexistant area "id: ALL, name: ALL, etc" that inside cameras contains all camera ids

JFer11: Okey, I understand ALL means all the areas. I change it. Now, ALL means all the cameras in an area called "All".

api/routers/areas.py
```diff
@@ -32,18 +32,34 @@ async def list_areas():
     }
 
 
+def get_all_cameras_ids():
+    ids_list = [camera['id'] for camera in get_cameras()]
+    return ",".join(ids_list)
+
+
+def all_areas():
```
mats-claassen, line 40: I'd rename this to `all_cameras_area`

api/routers/metrics/metrics.py
```diff
@@ -20,20 +20,29 @@ def get_cameras(cameras: str) -> Iterator[str]:
...
+
 def validate_camera_existence(camera_id: str):
     dir_path = os.path.join(os.getenv("SourceLogDirectory"), camera_id, "objects_log")
     if not os.path.exists(dir_path):
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"Camera with id '{camera_id}' does not exist")
 
 
 def get_areas(areas: str) -> Iterator[str]:
-    if areas and not ALL_AREAS in areas.upper():
+    if ALL_AREAS in areas.upper():
```
mats-claassen, line 35: Is this a list of ids? Could an id be `ALL...`? I would first split by `,` and then check if `ALL` is in there

api/routers/areas.py
JFer11: Changed!

api/routers/metrics/metrics.py
JFer11: Yes, you are right! Thanks!

mats-claassen (approved):

api/routers/areas.py
```diff
@@ -31,11 +32,34 @@ async def list_areas():
...
+    # Returns information about all the cameras in one area.
+    return {
+        "violation_threshold": -1,
+        "notify_every_minutes": -1,
+        "emails": "",
+        "enable_slack_notifications": False,  # "N/A"
+        "daily_report": False,  # "N/A"
+        "daily_report_time": "N/A",
+        "occupancy_threshold": -1,
+        "id": "All",
```
pgrill, line 50: Maybe we can use here the constant `ALL_AREAS`

renzodgc (approved):

api/routers/areas.py
JFer11: Great!

pgrill (approved):

PR #154, opened 2021-03-25
Occupancy rules

mats-claassen:
This PR adds the logic for multiple time based occupancy threshold rules.

It introduces a breaking change in that areas and video source objects are now classes. This means we cannot use the dict functions on it apart from get values like in a dict (which has been implemented)
For example: `if ... in area` or `area["something"] = ...` won't work anymore, but `asdf = area['id']` will.

Please review whether you can think of any place where this oucld break (I tested the reports: daily, hourly, live and mailing, not slack but that looks ok)

For calculating the daily threshold I took the max of the hourly thresholds for that day. This could be implemented using the default threshold but that is currently not available in the daily metrics function


libs/entities/occupancy_rule.py
```diff
@@ -0,0 +1,14 @@
...
+
+class OccupancyRule:
+
+    def __init__(self, json_value: dict):
+        self.days = json_value["days"]
+        start_time = json_value["start_hour"].split(":")
+        self.start_hour = time(int(start_time[0]), int(start_time[1]))
+        finish_time = json_value["finish_hour"].split(":")
+        self.finish_hour = time(int(finish_time[0]), int(finish_time[1]))
+        self.occupancy_threshold = json_value["max_occupancy"]
```
pgrill, line 11: Maybe we can include the validation  `start_time < end_time`.

libs/utils/utils.py
```diff
@@ -12,3 +14,8 @@ def validate_file_exists_and_is_not_empty(file_path):
 
 def is_list_recursively_empty(l):
     return all(is_list_recursively_empty(i) if isinstance(i, list) else False for i in l)
+
+def config_to_boolean(value):
+    if value.lower() not in BOOLEAN_STATES:
+        raise ValueError("Invalid boolean value")
+    return BOOLEAN_STATES[value.lower()]
```
pgrill, line 21: Can we add the newline? (Otherwise, my local linter will warning me :joy: )

libs/entities/occupancy_rule.py
```diff
@@ -0,0 +1,14 @@
...
+
+    def date_is_included(self, date: datetime):
+        return self.days[date.weekday()] and self.start_hour <= date.time() and self.finish_hour > date.time()
```
renzodgc, line 14:
```suggestion
        return self.days[date.weekday()] and self.start_hour <= date.time() and self.finish_hour > date.time()

```

libs/metrics/base.py
```diff
@@ -204,9 +204,9 @@ def compute_live_metrics(cls, config, live_interval):
             report_file_exists = os.path.isfile(live_report_csv)
             if not os.path.isfile(today_entity_csv):
                 return
-            entity["base_directory"] = entity_directory
             entries_in_interval = int(live_interval * 60 / get_source_logging_interval(config))
             live_data = cls.generate_live_csv_data(config, today_entity_csv, entity, entries_in_interval)
+            assert len(live_data) == len(csv_headers), "Row element count not the same as header count!!"
```
renzodgc, line 209: Why is this here? To indicate logs should be deleted?

libs/utils/utils.py
renzodgc:
```suggestion
    return BOOLEAN_STATES[value.lower()]

```

libs/entities/occupancy_rule.py
pgrill: I think that for this version we can split the range into two

mats-claassen: Yeah we agreed with FE that time ranges are only valid during the day. I have just added the validations with the endpoints

libs/metrics/base.py
mats-claassen: I found a bug where we were adding too more items to the `live_data` than headers. As we manually take the first N items from `live_data` where N = len(headers) there would be no exception, but the final data was wrong. This is to avoid such errors in future

pgrill (approved):

api/models/occupancy_rule.py
```diff
@@ -0,0 +1,102 @@
...
+    __root__: List[AreaOccupancyRule]
+
+    def __iter__(self):
+        return iter(self.__root__)
+
+    def __getitem__(self, item):
+        return self.__root__[item]
+
+    @validator('__root__')
+    def no_overlap_please(cls, the_list):
```
renzodgc, line 77: Shouldn't it be something like "rules_do_not_overlap"?

api/models/occupancy_rule.py
```diff
@@ -0,0 +1,102 @@
...
+        objs = [AreaOccupancyRule.parse_obj(v) for v in json_value["occupancy_rules"]]
+        return OccupancyRuleListDTO.parse_obj(objs)
+
+
+def do_overlap(a: AreaOccupancyRule, b: AreaOccupancyRule):
+    for d in range(7):
+        if a.days[d] and b.days[d]:
+            # sorry for the nots
+            if not (not date_before(b.start_time, a.finish_time, strict=True) or
+                    not date_before(a.start_time, b.finish_time, strict=True)):
```
renzodgc, line 100: This is the same as:

`if date_before(b.start_time, a.finish_time, strict=True) and date_before(a.start_time, b.finish_time, strict=True)
`
Right? unless I'm misinterpreting https://en.wikipedia.org/wiki/De_Morgan%27s_laws

api/models/occupancy_rule.py
mats-claassen: I could call it validate_overlaps, yes...

PR #157, opened 2021-03-30
In out estimate occupancy

renzodgc:
- Estimated latest, average and max occupancy
- Fixed boundary in/out summary in hourly report


libs/metrics/base.py
```diff
@@ -169,6 +166,9 @@ def compute_daily_metrics(cls, config):
                 if not report_file_exists:
                     writer.writeheader()
                 row = {"Date": yesterday}
+                print(row)
```
mats-claassen, line 169: Please remove these prints

libs/metrics/in_out.py
```diff
@@ -207,11 +208,85 @@ def get_weekly_report(cls, entities: List[str], number_of_weeks: int = 0,
...
+
+        # Normalize in_out:
+        return (int(total_in >= total_out and total_in > 0), int(total_out >= total_in and total_out > 0))
+
+    @classmethod
+    def _process_hourly_segments(
+        cls, hourly_objects_logs, latest_estimated_occupancy, boundaries, # input
+        hour_in, hour_out, hour_balance, summary_report # output
+    ):
+        logger.info(summary_report)
```
mats-claassen, line 251: This does not seem necessary. Same below and if you think it is right to remove, also remove the imports etc.

renzodgc: Yes, I thought I had them removed

libs/metrics/base.py
```diff
@@ -232,14 +229,14 @@ def get_hourly_report(cls, entities: List[str], report_date: date) -> Dict:
             file_path = os.path.join(reports_directory, f"report_{report_date}.csv")
             if os.path.exists(file_path):
                 df = pd.read_csv(file_path)
-                for header in cls.csv_headers:
+                for index, header in enumerate(cls.csv_headers):
```
pgrill, line 232: Why do you need the index here?

libs/metrics/in_out.py
```diff
@@ -207,11 +205,82 @@ def get_weekly_report(cls, entities: List[str], number_of_weeks: int = 0,
...
+
+def _read_estimated_latest_occupancy(in_out_file_path):
+    if os.path.exists(in_out_file_path):
+        with open(in_out_file_path, "r") as in_out_file:
+            latest_entry = deque(csv.DictReader(in_out_file), 1)
+            if len(latest_entry) != 0:
+                if "Time" not in latest_entry[0] or datetime.strptime(latest_entry[0]["Time"], "%Y-%m-%d %H:%M:%S").date() == datetime.today().date():
+                    return int(latest_entry[0]["EstimatedLatestOccupancy"])
+                else:
+                    return 0
```
pgrill, line 282: As you are using partial returns, you can simplify the if to:
```
if len(latest_entry) != 0:
    if "Time" not in latest_entry[0] or datetime.strptime(latest_entry[0]["Time"], "%Y-%m-%d %H:%M:%S").date() == datetime.today().date():
        return int(latest_entry[0]["EstimatedLatestOccupancy"])
# else
return 0
```

mats-claassen (approved):

PR #158, opened 2021-04-05
Get areas all

JFer11:
Various changes and fixes so that the get_areas () function returns a new area called "ALL" that points to all existing cameras. This area is a special area because it is not in the configuration file.


api/processor_api.py
```diff
@@ -72,6 +73,7 @@ def create_fastapi_app(self):
         app.include_router(config_router, prefix="/config", tags=["Config"], dependencies=dependencies)
         app.include_router(cameras_router, prefix="/cameras", tags=["Cameras"], dependencies=dependencies)
         app.include_router(areas_router, prefix="/areas", tags=["Areas"], dependencies=dependencies)
+        app.include_router(area_all_router, prefix="/all_areas", tags=["Areas"], dependencies=dependencies)
```
pgrill, line 76: I think that it should be called `/area_all`, otherwise it looks like the endpoint returns all the areas.

api/routers/area_all.py
```diff
@@ -0,0 +1,82 @@
...
+
+    if not os.path.exists(config_path):
+        # Create the file with if necessary
+        with open(config_path, 'x') as outfile:
+            json.dump(json_content, outfile)
+    else:
+        # If file exists, we have to modify the content in: "global_area_all".
+        with open(config_path, "r") as file:
+            file_content = json.load(file)
+
```
pgrill, line 73: I think that we can simplify the if.
Somrhing like 
```
if not os.path.exists(config_path):
    # Create the file with if necessary
    with open(config_path, 'w') as outfile:
        json.dump(json_content, {"global_area_all": {}})

with open(config_path, "r") as file:
    file_content = json.load(file)

file_content["global_area_all"] = json_content["global_area_all"]

with open(config_path, "w") as file:
    json.dump(file_content, file)
```

libs/entities/area.py
```diff
@@ -25,6 +26,45 @@ def __init__(self, config_section: dict, section_title: str, send_email_enabled:
             self.should_send_slack_notifications = False
         self.load_occupancy_rules()
 
+    @classmethod
+    def set_global_areas(cls, is_email_enabled, is_slack_enabled, config_dir, area_logs_dir, cameras_list):
```
pgrill, line 30: Should call it `set_global_area`. (without the s)

JFer11: Yes, done!

api/processor_api.py
JFer11: Yes, you are right!

api/routers/area_all.py
JFer11: Great idea!

api/tests/data/mocked_data/data/processor/config/areas/5.json
```diff
@@ -0,0 +1 @@
+{"occupancy_rules": [{"days": [1, 1, 0, 0, 0, 1, 1], "start_hour": "12:00", "finish_hour": "22:00", "max_occupancy": 12}, {"days": [1, 1, 0, 0, 0, 1, 1], "start_hour": "22:00", "finish_hour": "23:00", "max_occupancy": 11}]}
```
mats-claassen, line 1: Do we want this file checked in?

libs/entities/area.py
```diff
@@ -25,6 +26,45 @@ def __init__(self, config_section: dict, section_title: str, send_email_enabled:
             self.should_send_slack_notifications = False
         self.load_occupancy_rules()
 
+    @classmethod
+    def set_global_area(cls, is_email_enabled, is_slack_enabled, config_dir, area_logs_dir, cameras_list):
+        config_path = os.path.join(config_dir, "ALL.json")
+        json_content = {
+            "global_area_all": {
+                "ViolationThreshold": -1,
```
mats-claassen, line 34: I think 0 is a better default for `ViolationThreshold`, `NotifyEveryMinutes` and `OccupancyThreshold`

JFer11: Changed!

api/tests/data/mocked_data/data/processor/config/areas/5.json
JFer11: No, I create a fixture for each test in test_area_occupancy_rules.py, to remove the file that those tests created when they were run in my local environment.

api/routers/areas.py
```diff
@@ -74,6 +78,8 @@ async def create_area(new_area: AreaConfigDTO, reboot_processor: Optional[bool]
     """
     Adds a new area to the processor.
     """
+    # TODO: We have to autogenerate the ID.
+    # TODO: Test what happens when we send invalid occupancy rules, for example: [{}] (server explodes)
```
mats-claassen, line 82: Maybe this happens because the fields are Optional. I think none of them is optional though so maybe removing that is the answer

api/routers/areas.py
```diff
@@ -113,16 +119,65 @@ async def create_area(new_area: AreaConfigDTO, reboot_processor: Optional[bool]
     return next((area for area in get_areas() if area["id"] == area_dict["Id"]), None)
 
 
+def modify_area_all(area_information):
+    """
+    Edits the configuration related to the area "ALL", an area that contains all cameras.
+    """
+    # Doubt: We force the user to send "cameras", since the input must be AreaConfig DTO. However, I don't think
```
mats-claassen, line 126: The user should not send `cameras`. Can we make that field optional?

api/routers/areas.py
```diff
@@ -113,16 +119,65 @@ async def create_area(new_area: AreaConfigDTO, reboot_processor: Optional[bool]
...
+    # this is entirely correct.
+
+    config = get_config()
+    config_directory = config_utils.get_area_config_directory(config)
+    config_path = os.path.join(config_directory, ALL_AREAS + ".json")
```
mats-claassen, line 131: You should use `config.get_area_config_path(ALL_AREAS)`

api/routers/areas.py
```diff
@@ -201,14 +259,19 @@ def get_area_occupancy_rules(area_id: str):
...
 
     with open(area_config_path, "r") as area_file:
         rules_data = json.load(area_file)
     return OccupancyRuleListDTO.from_store_json(rules_data)
 
 
 def set_occupancy_rules(area_id: str, rules):
+    """
+    Doubt:
+    I don't think these functions are called correctly in the above endpoints.
```
mats-claassen, line 272: I think rules should be a `OccupancyRuleListDTO` if set. What might be wrong is the default value of that field in AreaConfigDTO which I think should either be None or an empty OccupancyRuleListDTO

api/routers/areas.py
```diff
@@ -230,8 +293,10 @@ def delete_area_occupancy_rules(area_id: str):
         with open(area_config_path, "r") as area_file:
             data = json.load(area_file)
     else:
+        # Doubt: I believe the line below does not have sense at all. why are we sending a 204 status code?
```
mats-claassen, line 296: This is copied from how other DELETE work but yeah, maybe there is a better status code for this

api/tests/app/test_area_occupancy_rules.py
```diff
@@ -68,15 +73,21 @@ def test_unitary_set_get_delete(self, config_rollback_areas):
         assert res1["occupancy_rules"][0]["max_occupancy"] == 12
         assert res2["occupancy_rules"][0]["max_occupancy"] == 100
 
-    def test_get_not_found(self, config_rollback_areas):
+        # Rollback the config file for area ID: 537
+        config_directory = config_utils.get_area_config_directory(get_config())
+        config_path = os.path.join(config_directory, str(area_id) + ".json")
```
mats-claassen, line 78: Here as well, prefer the helper function

api/tests/data/mocked_data/data/processor/config/areas/ALL.json
```diff
@@ -1 +1 @@
-{"global_area_all": {"ViolationThreshold": -1, "NotifyEveryMinutes": -1, "Emails": "", "EnableSlackNotifications": "false", "DailyReport": "false", "DailyReportTime": "N/A", "OccupancyThreshold": -1, "Id": "ALL", "Name": "ALL"}}
\ No newline at end of file
+{"global_area_all": {"ViolationThreshold": 0, "NotifyEveryMinutes": 0, "Emails": "", "EnableSlackNotifications": "False", "DailyReport": "False", "DailyReportTime": "06:00", "OccupancyThreshold": 0, "Id": "ALL", "Name": "ALL"}, "occupancy_rules": []}
```
mats-claassen, line 1: Maybe the booleans should be booleans and not strings?

libs/config_engine.py
```diff
@@ -185,6 +186,9 @@ def get_area_all(self):
         area_all = next(area for area in areas if area.id == ALL_AREAS)
         return area_all
 
+    def get_area_config_path(self, area_id):
```
mats-claassen, line 189: Wonder why this appears as changed here. This is already on master

api/routers/areas.py
JFer11: At first I thought the field must be mandatory because cameras are used by various functions, and sending None will throw an exception. However, I think if we set a default "" when no camera is sent, everything will work fine. I will do it.

api/routers/areas.py
JFer11: Done!

JFer11: Done!

api/routers/areas.py
JFer11: I mean, the status code here doesn't matter because since the second parameter is False, the handle_response function will always throw a 500 error. I'll remove the 204 status code to avoid confusion.

api/routers/areas.py
mats-claassen: It seems they are not optional so 🤷🏻‍♂️

api/tests/app/test_area_occupancy_rules.py
JFer11: Done!

libs/config_engine.py
JFer11: IDK maybe because I resolved conficlts?

api/tests/data/mocked_data/data/processor/config/areas/ALL.json
JFer11: Of course, I change the process of creating and modifying the ALL.json file and now the boolean values are stored.

api/routers/areas.py
JFer11: What I was trying to explain is that:
set_occupancy_rules is called with two parameters.
An identification (chain) and rules. If we pay attention to where we are using the rules parameter within the function, we can notice that we call its .to_store_json () method.

```
def set_occupancy_rules(area_id: str, rules):
    """
    Doubt:
    I don't think these functions are called correctly in the above endpoints.
    Rules are lists and lists do not have .to_store_json()
    """
    area_config_path = get_config().get_area_config_path(area_id)
    Path(os.path.dirname(area_config_path)).mkdir(parents=True, exist_ok=True)

    if os.path.exists(area_config_path):
        with open(area_config_path, "r") as area_file:
            data = json.load(area_file)
    else:
        data = {}

    with open(area_config_path, "w") as area_file:
        data["occupancy_rules"] = rules.to_store_json()["occupancy_rules"]
        json.dump(data, area_file)
```

However, when we call the function, "rules" parameter is a list of AreaOccupancyRule, but it is a list. And it does not have the function .to_store_json()
```
    occupancy_rules = edited_area.occupancy_rules
    del edited_area.occupancy_rules

    area_dict = map_to_config_file_format(edited_area)
    config_dict[f"Area_{index}"] = area_dict
    success = update_config(config_dict, reboot_processor)

    if occupancy_rules:
        set_occupancy_rules(edited_area.id, occupancy_rules)
    else:
        delete_area_occupancy_rules(area_id)
```

mats-claassen (approved):

PR #160, opened 2021-04-06
Merge endpoints for area and occupancy

mats-claassen:
Also add a small fix for In/Out with latest changes to entities classes

### DIFF
--- a/api/routers/areas.py
+++ b/api/routers/areas.py
@@ -101,18 +101,7 @@ async def create_area(new_area: AreaConfigDTO, reboot_processor: Optional[bool] 
     success = update_config(config_dict, reboot_processor)
 
     if occupancy_rules:
-        area_config_path = get_config().get_area_config_path(new_area.id)
-        Path(os.path.dirname(area_config_path)).mkdir(parents=True, exist_ok=True)
-
-        if os.path.exists(area_config_path):
-            with open(area_config_path, "r") as area_file:
-                data = json.load(area_file)
-        else:
-            data = {}
-
-        with open(area_config_path, "w") as area_file:
-            data["occupancy_rules"] = occupancy_rules.to_store_json()["occupancy_rules"]
-            json.dump(data, area_file)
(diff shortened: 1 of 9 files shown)


api/routers/areas.py
```diff
@@ -132,9 +151,29 @@ async def edit_area(area_id: str, edited_area: AreaConfigDTO, reboot_processor:
...
+
+        if os.path.exists(area_config_path):
+            with open(area_config_path, "r") as area_file:
+                data = json.load(area_file)
+        else:
+            data = {}
+
+        with open(area_config_path, "w") as area_file:
+            data["occupancy_rules"] = occupancy_rules.to_store_json()["occupancy_rules"]
+            json.dump(data, area_file)
```
renzodgc, line 173: Can we encapsulate this into a method?

api/routers/areas.py
```diff
@@ -87,21 +87,40 @@ async def create_area(new_area: AreaConfigDTO, reboot_processor: Optional[bool]
...
+
     if not success:
         return handle_response(area_dict, success, status.HTTP_201_CREATED)
 
     area_directory = os.path.join(os.getenv("AreaLogDirectory"), new_area.id, "occupancy_log")
     Path(area_directory).mkdir(parents=True, exist_ok=True)
 
+    # known issue: Occupancy rules not returned
     return next((area for area in get_areas() if area["id"] == area_dict["Id"]), None)
```
pgrill, line 124: Maybe we can create a ticket to fix it.
I think that it can be fixed easily adding  in the `get_areas` function a parameter called `include_occupancy_rules=False`

renzodgc: It would be fixed if we migrate from the .ini to a .json for all areas, right? That's what we discussed on the daily

mats-claassen: This `get_areas` function is local in this file, not the one from config_engine. I don't think we should add that logic here as well.

I can create a ticket for it

pgrill (approved):

renzodgc (approved):

PR #163, opened 2021-04-14
deploy infrastructure with terraform in AWS

agustinasierralima:
@pgrill @mats-claassen

### DIFF
--- a/terraform/terraform.tfvars
+++ b/terraform/terraform.tfvars
@@ -1,0 +1,31 @@
+# Generics
+region = "us-east-1"
+ 
+tags = [{ Project = "lanthorn",State = "Deploy EC2 with nvidia drivers and docker installed"}]
+
+
+# Networking
+vpc = [{ name = "lanthorn_vpc", cidr_block = "10.0.0.0/16" , instance_tenancy = "default", enable_dns_hostnames = true, enable_dns_support = true}]
+
+subnet_prefix = [{ name = "langthron_module_subnet", cidr_block = "10.0.1.0/24", availability_zone = "us-east-1a", map_public_ip_on_launch = true }]
+
+security_group = [{ name = "langthron_module_security_group" }]
+
+route_table = [{ name = "langthron_module_route_table", cidr_block = "0.0.0.0/0", ipv6_cidr_block = "::/0" }]
+
+network_interface = [{ name = "langthron_module_network_interface", private_ips = ["10.0.1.50"], device_index = 0 }]
+
+
+# EC2
+ec2name = "langthron-smart-social-distancing"
+
+key_name = "lanthornkey"
+
+size = "g4dn.xlarge"
+
+root_disk = [{ volume_size = "100",volume_type = "gp2"}]
+
+amiid = "ami-012e9f6aa634f84f8"
+
+# To be run after creation
+user_data_path = "user_data.tmpl"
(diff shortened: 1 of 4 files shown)


terraform/terraform.tfvars
```diff
@@ -0,0 +1,31 @@
+# Generics
+region = "us-east-1"
+ 
+tags = [{ Project = "lanthorn",State = "Deploy EC2 with nvidia drivers and docker installed"}]
+
+
+# Networking
+vpc = [{ name = "lanthorn_vpc", cidr_block = "10.0.0.0/16" , instance_tenancy = "default", enable_dns_hostnames = true, enable_dns_support = true}]
+
+subnet_prefix = [{ name = "langthron_module_subnet", cidr_block = "10.0.1.0/24", availability_zone = "us-east-1a", map_public_ip_on_launch = true }]
```
mats-claassen, line 10: langthron_module_subnet -> lanthorn...

I wonder about all of these ids. They only work for a specific account right? So we will have to mention that they must be changed

terraform/user_data.tmpl
```diff
@@ -0,0 +1,34 @@
...
+sudo yum install nvidia* -y
+
+distribution=$(. /etc/os-release;echo $ID$VERSION_ID) \
+   && curl -s -L https://nvidia.github.io/nvidia-docker/$distribution/nvidia-docker.repo | sudo tee /etc/yum.repos.d/nvidia-docker.repo
+
+sudo yum clean expire-cache
+
+sudo systemctl restart docker
+
+sudo docker run --rm --gpus all nvidia/cuda:11.0-base nvidia-smi
```
mats-claassen, line 22: I don't think this is necessary

terraform/user_data.tmpl
```diff
@@ -0,0 +1,34 @@
...
+
+#Api
+sudo yum install git -y
+git clone https://github.com/neuralet/smart-social-distancing.git
+cd smart-social-distancing
+git fetch --tags
+git checkout $(git tag | tail -1)
+./download-x86-openpifpaf-model.sh
+
+#docker run -d -it --gpus all --restart unless-stopped --name smart-social-distancing -p 443:8000 -v $PWD/data:/repo/data -v $PWD/certs:/repo/certs -v $PWD/config-x86-gpu-tensorrt.ini:/repo/config-x86-gpu-tensorrt.ini -e TZ=`./timezone.sh` neuralet/smart-social-distancing:latest-x86_64_gpu_tensorrt
```
mats-claassen, line 32: Should this be uncommented?

terraform/terraform.tfvars
```diff
@@ -0,0 +1,31 @@
+# Generics
```
pgrill, line 1: I think that this file should be completed by the customer with the values he wants.
Maybe we can keep if empty and show examples using comments.

For example:
```
region =  # Example: "us-east-1"
```

What do you think?

sebaxmartlabs: tfvars is the file where the customer will add his own values for the variables defined in it. I think the use case is included in the current set-up.

pgrill: Ok, so in that case, I prefer to have it empty, so if the customer forgot to complete it, the script will through an error.

mats-claassen (approved):

pgrill (approved):

PR #170, opened 2021-05-25
Support past video processing

pgrill:


libs/cv_engine.py
```diff
@@ -115,9 +125,21 @@ def __process(self, cv_image):
             self.log_detail["Tracker"].append(tracker_time)
         return cv_image, tmp_objects_list, post_processing_data
 
-    def process_video(self, video_uri):
+    def process_video_file(self, video_uri, video_date: str = None):
         input_cap = cv.VideoCapture(video_uri)
-        fps = max(25, input_cap.get(cv.CAP_PROP_FPS))
+        fps = input_cap.get(cv.CAP_PROP_FPS)
```
mats-claassen, line 130: this is overridden below so you can remove this line

libs/loggers/source_loggers/file_system_logger.py
```diff
@@ -32,7 +32,8 @@ def save_screenshot(self, cv_image):
 
     def log_objects(self, objects, violating_objects, violating_objects_index_list, violating_objects_count,
                     detected_objects_cout, environment_score, time_stamp, version):
-        file_name = str(date.today())
+        
+        file_name = time_stamp[0: 10]
```
mats-claassen, line 36: Is this always safe to do?

libs/cv_engine.py
pgrill: Removed

libs/loggers/source_loggers/file_system_logger.py
pgrill: Yes, the time_stamp parameter always receives the result of apply `strftime("%Y-%m-%d %H:%M:%S")` to a datetime.

mats-claassen (approved):

PR #174, opened 2021-05-31
Add Dwell time metric

mats-claassen:


constants.py
```diff
@@ -10,3 +10,4 @@
 SOCIAL_DISTANCING = "social-distancing"
 FACEMASK_USAGE = "facemask-usage"
 IN_OUT = "in-out"
+DWELL_TIME = "dwell-time"
```
pgrill, line 13: Can we add the end of line?

libs/metrics/base.py
```diff
@@ -325,6 +328,7 @@ def get_daily_report(cls, entities: List[str], from_date: date, to_date: date) -
         base_directory = cls.get_entity_base_directory()
         date_range = pd.date_range(start=from_date, end=to_date)
         base_results = {}
+        daily_headers = [h for h in cls.csv_headers if h not in cls.ignored_headers]
```
pgrill, line 331: Maybe you can include the following property
```
@property
def report_headers(self):
         return [h for h in cls.csv_headers if h not in cls.ignored_headers]
```
and use it instead fd repeating the same logic.

pgrill (approved):

renzodgc (approved):

